PR #1, opened 2019-01-14
Helm repo sync and lint

### LINKED ISSUE #2: Setup Circle CI/CD pipeline for charts repo
Adapt community charts repo scripts for our lint/test/deploy steps

sudermanjr:
Fixes #2 

Touches on #4 as well.

### DIFF
--- a/scripts/repo-sync.sh
+++ b/scripts/repo-sync.sh
@@ -85,7 +85,7 @@ sync_repo() {
         # Move updated index.yaml to sync folder so we don't push the old one again
         mv -f "$sync_dir/index.yaml" "$index_dir/index.yaml"
 
-        aws s3 sync "$sync_dir" "$bucket"
+        aws s3 sync --exclude index.yaml "$sync_dir" "$bucket"
 
         # Make sure index.yaml is synced last
         aws s3 cp "$index_dir/index.yaml" "$bucket"
--- a/scripts/repo-sync.sh
+++ b/scripts/repo-sync.sh
@@ -1,0 +1,106 @@
+#!/usr/bin/env bash
+
+# Copyright 2018 The Kubernetes Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+readonly HELM_URL=https://storage.googleapis.com/kubernetes-helm
+readonly HELM_TARBALL=helm-v2.12.2-linux-amd64.tar.gz
+readonly STABLE_REPO_URL=https://charts.reactiveops.com/stable/
+readonly INCUBATOR_REPO_URL=https://charts.reactiveops.com/incubator/
+readonly S3_BUCKET_STABLE=s3://charts.reactiveops.com/stable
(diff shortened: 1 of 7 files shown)


stable/rbac-manager/CODEOWNERS
```diff
@@ -1 +1 @@
-* @robscott @davekonopka @sudermanjr
```
sudermanjr, line 1: FYI, I left only the chart maintainers here, I just happen to be both in this case. _Dave will be included by the top-level.

I left my name here in case I ever stop being a CODEOWNER of the entire repo.

stable/rbac-manager/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: rbac-manager
-version: 0.1.0
```
sudermanjr, line 2: Had to bump this since the chart folder was changing.  Maybe there's a way around that.....

scripts/repo-sync.sh
```diff
@@ -0,0 +1,106 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+readonly HELM_URL=https://storage.googleapis.com/kubernetes-helm
+readonly HELM_TARBALL=helm-v2.12.2-linux-amd64.tar.gz
+readonly STABLE_REPO_URL=https://charts.reactiveops.com/stable/
+readonly INCUBATOR_REPO_URL=https://charts.reactiveops.com/incubator/
+readonly S3_BUCKET_STABLE=s3://charts.reactiveops.com/stable
+readonly S3_BUCKET_INCUBATOR=gs://charts.reactiveops.com/incubator
```
davekonopka: Looks like a `gs` reference needs to be changed to `s3`.

scripts/repo-sync.sh
```diff
@@ -0,0 +1,106 @@
...
+    done
+
+    if helm repo index --url "$repo_url" --merge "$index_dir/index.yaml" "$sync_dir"; then
+        # Move updated index.yaml to sync folder so we don't push the old one again
+        mv -f "$sync_dir/index.yaml" "$index_dir/index.yaml"
+
+        aws s3 sync "$sync_dir" "$bucket"
+
+        # Make sure index.yaml is synced last
+        aws s3 cp "$index_dir/index.yaml" "$bucket"
```
davekonopka, line 91: Seems like the index.yaml will be moved as part of the `aws s3 sync` above. Maybe the index shouldn't be moved into the sync directory or an exclude should be added to the sync to keep it out of the bucket until last?

scripts/repo-sync.sh
```diff
@@ -0,0 +1,106 @@
...
+
+readonly HELM_URL=https://storage.googleapis.com/kubernetes-helm
+readonly HELM_TARBALL=helm-v2.12.2-linux-amd64.tar.gz
+readonly STABLE_REPO_URL=https://charts.reactiveops.com/stable/
+readonly INCUBATOR_REPO_URL=https://charts.reactiveops.com/incubator/
+readonly S3_BUCKET_STABLE=s3://charts.reactiveops.com/stable
+readonly S3_BUCKET_INCUBATOR=gs://charts.reactiveops.com/incubator
+
+main() {
+    setup_helm_client
```
davekonopka, line 29: Helm is already installed in the helmpack chart-testing image. Do we need to install a specific version again here? Not a big deal either way.

davekonopka (commented):
Left a few comments/questions inline. Overall looks good.

scripts/repo-sync.sh
sudermanjr: Yeah, I had that thought as well.  I think it's worth keeping in case we want to update it to a newer version.  This was also done in the main helm repo, and I think this is why.

scripts/repo-sync.sh
sudermanjr: I was wondering about that myself.  The helm repo had this in there to make sure it synced last, but I'm honestly not sure why.

scripts/repo-sync.sh
sudermanjr: Fixed.

davekonopka (approved):
Changes look good.

PR #6, opened 2019-01-14
Changing circle to sync on merge to master

sudermanjr:


.circleci/config.yml
```diff
@@ -40,12 +40,13 @@ workflows:
...
-            branches:
-              only:
-                - master
+    filters:
+      branches:
+        only:
+          - master
     jobs:
-      - sync
+      sync:
```
davekonopka: Does `jobs` value need to be an array?

sudermanjr: Yeah.  I just simplified it some more and fixed that.

davekonopka (approved):
👍

PR #7, opened 2019-01-15
Add e2e tests

sudermanjr:
Spin up a kind cluster, install helm/tiller, then install all of the helm charts with default values.


sudermanjr:
I think this is ready to go.  E2e tests aren't perfect yet, but they will fire and are not required for merging PRs.  This means we can tweak them as we go along.

davekonopka:
@sudermanjr Works well. I ran through the script locally and it ran cleanly for me.

A few things I noticed that are probably covered in "aren't perfect yet":

* The tests seem to clone the repo and run against master rather than the active branch.
* The aws-iam-authenticator doesn't seem to bring up a pod cleanly but didn't fail the ct test run. Less worried about the specific failure vs catching any failures.

sudermanjr:
Yeah, good catches, I noticed that as well.  I believe the helm chart testing doesn't care so much that the pods run as much as it cares that helm is able to install.  The environment will probably never be able to handle running all the pods that the different charts are going to run since they are all expecting specific external/internal factors, and some will require config that we just can't give it.

I realized the branch thing, which is annoying.  I'm going to have to pass the branch to the container so that it can checkout the right version.

sudermanjr:
I added the branch name to the script and the circle config.  I think it's working well and was easier than expected.

davekonopka (approved):
🎆 LGTM. Nice work.

PR #8, opened 2019-01-17
Adding rbac-manager chart to stable

sudermanjr:


stable/rbac-manager/CODEOWNERS
```diff
@@ -0,0 +1 @@
+* robscott sudermanjr
```
davekonopka: Do the usernames work without the `@` prefix?

davekonopka (approved):
👍

stable/rbac-manager/CODEOWNERS
sudermanjr: no, I don't think so

PR #10, opened 2019-01-17
Adding capsize

sudermanjr:


incubator/capsize/capsize.py
```diff
@@ -0,0 +1,92 @@
...
+"""
+import os
+import json
+from datetime import datetime
+import logging
+import pytz
+from kubernetes import client, config
+
+# Config from environment variables
+namespace = os.environ.get("KUBE_NAMESPACE", "istio-system")
```
davekonopka: Minor point but having actual istio-ingressgateway pods as a default further down in the script layer feels risky or at least potentially confusing. The chart defaults override this but I'd rather see these defaults match up with the chart to avoid causing unexpected restarts.

davekonopka (approved):
Looks good to me. One minor comment left inline but it's not a blocker.

incubator/capsize/capsize.py
sudermanjr: I just removed the defaults.  The chart specifies some fake defaults, so that should be good enough.  I would rather the script error out rather than restart an unintended pod.

sudermanjr: In addition, I added more checking to ensure the required variables are set and handled the errors gracefully.

PR #13, opened 2019-01-23
Adding chart to create custom iptables rules for memorystore

### LINKED ISSUE #14: e2e tests all charts every time
When making a branch with a new chart, the e2e tests run on every single chart.  This makes the test unnecessarily long.  It should just run for the chart in the changeset.

sudermanjr:
Also fixes #14 by removing the `--all`

### DIFF
--- a/incubator/custom-iptables/.helmignore
+++ b/incubator/custom-iptables/.helmignore
@@ -1,0 +1,21 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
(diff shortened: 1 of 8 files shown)


mjhuber (approved):
LGTM - this will be super useful for use with MemoryStore!

davekonopka (approved):
Looks good to me too.

I haven't looked closely at the script. How do you feel about how well it handles misconfiguration? If a weird config map gets posted for some reason is it likely to melt down iptables?

sudermanjr:
> Looks good to me too.
> 
> I haven't looked closely at the script. How do you feel about how well it handles misconfiguration? If a weird config map gets posted for some reason is it likely to melt down iptables?

I haven't either.  Perhaps I'm being too trusting of the GKE documentation.

PR #20, opened 2019-01-24
Adding ro-cert-manager to incubator

sudermanjr:
Added the ability to install cert-manager as a sub-chart.  This will make it so we can just install one chart.  Bumped to version 2.0.0 to cover this major change.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ jobs:
             go get sigs.k8s.io/kind
             cp /go/bin/kind /binaries/kind
       - save_cache:
-          key: v1-binaries
+          key: v2-binaries
           paths:
             - /binaries
   lint-scripts:
@@ -59,7 +59,7 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - v1-binaries
+            - v2-binaries
       - setup_remote_docker
       - *setup_environment
       - run:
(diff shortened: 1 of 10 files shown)


.circleci/config.yml
```diff
@@ -26,7 +26,7 @@ jobs:
             go get sigs.k8s.io/kind
             cp /go/bin/kind /binaries/kind
       - save_cache:
-          key: v1-binaries
+          key: v2-binaries
```
endzyme, line 29: is this change directly pertinent to the Pull Request of adding ro-cert-manager to incubator? Or is it included to not have yet another PR for a small change? (Trying to see if this should be under the microscope or is correlated to the files in the rest of the PR)

endzyme (approved):
Assuming the circleci change is non-coupled as all as assuming there haven't been any significant changes in the migration from the other repo, :+1: :+1:

.circleci/config.yml
sudermanjr: I discovered during testing that a new version of kind was released since the last build, and this is intended to force a rebuild of the kind binary.  Circle doesn't invalidate that cache very often (which in this case is just the kind binary).

So it is just avoiding an additional PR :-D

sudermanjr:
The only significant change is the inclusion of cert-manager as a sub-chart.  This was necessary in order for e2e testing to function.  I also believe it's a nice addition to the usability of the chart.

endzyme:
Love it - looks great :+1:

davekonopka (approved):
LGTM

PR #21, opened 2019-01-25
Add helm-release-pruner chart to incubator

davekonopka:


sudermanjr (approved):
Passes tests and the chart worked before, so LGTM.

PR #29, opened 2019-02-06
Move logentries chart to incubator

davekonopka:


sudermanjr (approved):
Cleaned up a couple little nitpicks and pushed.  LGTM

PR #31, opened 2019-02-08
[incubator/fluentd] Papertrail liveness et al

coreypobrien:
The original goal was to add a liveness probe for papertrail. In the process a whole lot of cleanup and improvement happened.

* Split config into separate files and use a common Kubernetes config (Remove tailing of salt, kubelet, docker, etc that were pulled from upstream configs)
* Common: Move verify_ssl to a shared setting since it is kubernetes-related not endpoint related (breaking)
* Common: Parameterize updateStrategy and default to rollingUpdate.maxUnavailable=10
* Common: Use nindent everywhere to make yaml more readable
* Common: Add checksum/config annotation to for Pod update on config change

* Papertrail: Add papertrail.extraArgs
* Papertrail: Force severity to info for all kube-* logs since 'I' and 'E' aren't valid
* Papertrail: Add livenessprobe to deal with upstream bug
* Papertrail: Disable FLUENTD_SYSTEMD_CONF since we don't have systemd access from the container


sudermanjr:
Since this chart has so many use-cases, would it be possible for you to add some additional testing values files?  We now have the ability to e2e test with `ci/*-values.yaml` files.  It would be nice to have one for each configuration so that we can test each one.

sudermanjr (changes requested):
This looks like a really nice cleanup and refactor.  Thanks!

Since this chart has so many use-cases, would it be possible for you to add some additional testing values files? We now have the ability to e2e test with `ci/*-values.yaml` files. It would be nice to have one for each configuration so that we can test each one.

coreypobrien:
CI files added! ♪┏(・o･)┛♪

sudermanjr:
Strange.  The test doesn't seem to have installed them all.

sudermanjr (approved):

dosullivan (approved):
Looks good to me. I like this approach to templating the configmap a lot more than what we had previously.

PR #38, opened 2019-02-13
alb ingress controller chart

mjhuber:
I added the alb-ingress-controller chart.  The community chart doesn't have a pdb or hpa, so I added that.

### DIFF
--- a/incubator/aws-alb-ingress-controller/templates/hpa.yaml
+++ b/incubator/aws-alb-ingress-controller/templates/hpa.yaml
@@ -1,0 +1,23 @@
+apiVersion: autoscaling/v1
+kind: HorizontalPodAutoscaler
+metadata:
+  name: {{ include "aws-alb-ingress-controller.fullname" . }}
+  labels:
+    app.kubernetes.io/name: {{ include "aws-alb-ingress-controller.name" . }}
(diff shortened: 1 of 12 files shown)


davekonopka:
I'm not sure how we should handle this being a fork of a community chart. At the very least we should probably add info to the readme that it is a fork.

incubator/aws-alb-ingress-controller/templates/hpa.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: autoscaling/v1
```
davekonopka, line 1: Does the controller app handle running more than one instance?

davekonopka:
@mjhuber Based on the upstream community chart feedback that this controller runs in leader/passive mode when you run multiple controllers the HPA is not going to be useful. I vote for getting your PR merged on the community chart and closing out this PR to skip adding the fork here.

mjhuber:
> @mjhuber Based on the upstream community chart feedback that this controller runs in leader/passive mode when you run multiple controllers the HPA is not going to be useful. I vote for getting your PR merged on the community chart and closing out this PR to skip adding the fork here.

i agree - closing

PR #40, opened 2019-02-15
Update helm release parsing / Support for tiller_namespace

mjhuber:
I updated the helm release parsing and added support for tiller in other namespaces.


incubator/helm-release-pruner/templates/cronjob.yml
```diff
@@ -33,6 +33,9 @@ spec:
             - mountPath: /usr/local/bin/prune-releases.sh
               name: config
               subPath: prune-releases.sh
+            env:
+            - name: TILLER_NAMESPACE
+              value: {{ .Values.tiller.namespace }}
```
davekonopka: Can you add a default value for the namespace in the values file?

mjhuber: Sure.  I added kube-system as a default.

incubator/helm-release-pruner/templates/rbac.yml
```diff
@@ -30,3 +30,25 @@ subjects:
 - kind: ServiceAccount
   name: {{ template "helm-release-pruner.fullname" . }}
   namespace: {{ .Release.Namespace }}
+---
```
davekonopka, line 33: I'd like to avoid giving the service account access to all namespaces to avoid accidentally deleting staging or other non-ephemeral deploys. That seems like a bad default for the chart to grant.

I had put the role binding in the app deploy to accomplish this originally: https://github.com/trueandco/trueandco-web/blob/master/deploy/ephemeral/helm-release-pruner.service_account.yml

Now that rbac-manager supports label based bindings I'd lean towards removing all bindings here and in the app deploy and add a label based privilege for the pruner service account in the reckoner rbac definitions file.

What would you think about that?

mjhuber: that sounds logical, as long as we're ok with having a dependency on a recent version of rbac-manager

mjhuber: I added an RbacDefinition thats disabled by default, and removed the namespace access role.

davekonopka (approved):
👍

sudermanjr:
Due to the changes in RBAC, is there actually an upgrade path between the old version and this version of the chart?  If not, I think we should up the major version and make a large note about the changes in the README.  Either way, I think maybe a note in the README about the change is worthwhile.

mjhuber:
> Due to the changes in RBAC, is there actually an upgrade path between the old version and this version of the chart? If not, I think we should up the major version and make a large note about the changes in the README. Either way, I think maybe a note in the README about the change is worthwhile.

Good point, I updated the readme and incremented the chart version.

sudermanjr (approved):
Awesome.  Thanks!

PR #42, opened 2019-02-20
Adding a demo app chart and a load generator chart

sudermanjr:
These I use frequently for demos and can be quite handy.


davekonopka (approved):
These look good to me. And handy to have.

PR #47, opened 2019-02-26
ECR Cleanup chart is stable.

sudermanjr:
Will also consider a PR to mention the chart as an install
method in the developer repo.


ejether (approved):
Chart looks fine but I'm pretty far from the operation to know if its the right thing or not.

davekonopka (approved):

PR #50, opened 2019-03-13
[incubator/fluentd] Fix duplicate system config for elasticsearch

coreypobrien:
The `<system>` section is in the global config so we need to remove this so we don't have a duplicate invalid config


sudermanjr (approved):
LGTM.

davekonopka (approved):
👍 LGTM

PR #51, opened 2019-03-13
custom-iptables: Make SLEEP_INTERVAL configurable

davekonopka:


incubator/custom-iptables/templates/daemonset.yaml
```diff
@@ -25,6 +25,9 @@ spec:
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           command: [ "sh", "-c", "/run.sh" ]
+          env:
+          - name: SLEEP_INTERVAL
+            value: "{{ .Values.sleepInterval }}"
```
sudermanjr: In my experience with helm and integer translation, `{{ .Values.sleepInterval | quote }}` works more consistently, but I haven't tried this method in a long time, so it might be fine.

sudermanjr (approved):

sudermanjr:
Approved pending tests pass

PR #53, opened 2019-03-20
updating livenessProbe for fluentd to match Corey's live patch

dosullivan:
Updating the fluentd chart to use a slightly more effective search pattern in it's livenessProbe. If any buffer files are found to be older than 5 minutes, it will immediately fail the livenessProbe.


sudermanjr (approved):
LGTM and passes tests.  I don't have the context for the actual change though.

PR #55, opened 2019-03-25
Adding an autospotting chart

### LINKED ISSUE #36: Move the autospotting chart to charts repo
We have an autospotting chart floating around that should be moved to incubator.

sudermanjr:
Fixes #36

### DIFF
--- a/incubator/autospotting/Chart.yaml
+++ b/incubator/autospotting/Chart.yaml
@@ -3,3 +3,6 @@ appVersion: "1.0"
 description: A chart to install autospotting.
 name: autospotting
 version: 0.1.0
+maintainers:
+  - name: @sudermanjr
+  - name: @coreypobrien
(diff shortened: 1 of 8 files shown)


incubator/autospotting/Chart.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+appVersion: "1.0"
```
davekonopka, line 2: Should this match our default image build version?

davekonopka (approved):
One minor comment inline. Otherwise 👍🏻

incubator/autospotting/Chart.yaml
sudermanjr: I figured I would go increment version numbers once the build of the container goes public and we have a latest version tag.

PR #57, opened 2019-03-27
[incubator/fluentd] Improving Elasticsearch Support

robscott:
This makes a number of changes to get Elasticsearch working with the latest official image. The defaults here are configured to assume AWS Elasticsearch config, but this should work well with any endpoints by overriding the default port and scheme. After experimenting with a number of options for config, I opted to stick with the config built into the image, and add any additional config in `conf.d` files [included by the default config](https://github.com/fluent/fluentd-kubernetes-daemonset/blob/master/templates/conf/fluent.conf.erb#L10).


sudermanjr (approved):
Awesome.  Very happy to not be hosting our own image of this anymore.

PR #58, opened 2019-03-28
Adding codeowners to every chart

sudermanjr:


sudermanjr:
I don't think it's necessary to bump all these versions for this change. @davekonopka if you agree, we can administrator merge

davekonopka (approved):
I'm fine with not bumping versions for this.

PR #60, opened 2019-04-02
[incubator/fluentd] Update ES and remove Loggly

coreypobrien:
- Remove Loggly since it isn't working, isn't used, and isn't using the standard images
- Add the custom elasticsearch config back in
- Switch everything to assume v0.12.43 images instead of a mix of 0.12 and 1.3
- Override the entrypoint to call fluentd directly and use --no-supervisor
- Since we're overriding the entrypoint, make `extraArgs` work by changing it to actual list of args
- Add `additional_filters_and_sources` to allow for any configuration to include additional fluentd configs


incubator/fluentd/templates/configmap.yaml
```diff
@@ -9,16 +9,15 @@ metadata:
     helm.sh/chart: {{ include "fluentd.chart" . }}
 apiVersion: v1
 data:
-  {{- if .Values.elasticsearch.enabled }}
-  custom.conf: |
-    {{ .Values.elasticsearch.customConf | nindent 4 }}
-  {{- else }}
   fluent.conf: |
     {{ tpl (.Files.Get "files/kubernetes.conf") . | nindent 4 }}
+    {{ .Values.additional_filters_and_sources | nindent 4 }}
```
davekonopka, line 14: Will it cause any issues if this outputs a blank line in the config when the value is empty?

davekonopka (approved):
Minor question inline. Otherwise makes sense to me. Once e2e passes: 👍

incubator/fluentd/templates/configmap.yaml
coreypobrien: Good question, but nope, its xml so indentation doesn't even really matter.

robscott (approved):
Looks good to me, thanks for the updates!

PR #61, opened 2019-04-02
[incubator/fluentd] Make ES buffer params configurable

coreypobrien:


incubator/fluentd/values.yaml
```diff
@@ -12,6 +12,12 @@ elasticsearch:
   port: 443
   scheme: https
   extraArgs: []
+  # This is set low to accomodate m4.large ES clusters that have a 10m client body limit
+  buffer_chunk_limit: 5m
```
dosullivan, line 16: maybe we could make a map in a future release to provide multiple defaults for `buffer_chunk_limit` and `buffer_queue_limit` based on instance type?

coreypobrien: Definitely 👍These are targeted toward minimums, but could be way more efficient if instance allowed for it.

robscott (approved):
Looks good to me, thanks!

davekonopka (approved):
👍🏻

PR #62, opened 2019-04-10
[incubator/fluentd] ES improvements and kubernetes_cluster

coreypobrien:
* Add ES settings for buffer_queue_full_action and request_timeout
* Drop all fluentd logs. When there are problems pushing output, generating errors that then need to also be pushed only componds the problem
* Require `cluster_name` to set the `kubernetes_cluster` field on all records
* Remove unused papertrail.cluster_name setting

### DIFF
--- a/incubator/fluentd/values.yaml
+++ b/incubator/fluentd/values.yaml
@@ -18,6 +18,8 @@ elasticsearch:
   buffer_queue_limit: 100
   flush_interval: 5s
   num_threads: 4
+  request_timeout: 20s
+  buffer_queue_full_action: exception
 
 ######################
 ##    Papertrail    ##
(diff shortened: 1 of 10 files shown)


incubator/fluentd/values.yaml
```diff
@@ -31,7 +33,6 @@ papertrail:
   enabled: false
   host: logs3.papertrailapp.com
   port: 12785
-  cluster_name:
```
dosullivan, line 34: Instead of removing the `papertrail.cluster_name` nested key entirely, shouldn't we break `cluster_name` out to be its own key as it is required?

sudermanjr (approved):
Assuming it passes tests, LGTM

incubator/fluentd/values.yaml
coreypobrien: Added a commented out version so we still require it but it is present in the values.yaml

PR #65, opened 2019-04-11
[incubator/fluentd] Fix k8s audit log processing

coreypobrien:
* Parse as JSON
* Fix default path but allow for other paths
* Pass responseObject and requestObject as strings to avoid having the whole k8s api spec as separate fields


sudermanjr (approved):
LGTM

robscott (approved):
Very cool, thanks for getting this in! Default path matches up with our Pentagon default.

PR #66, opened 2019-04-22
Ignore kube-bench controls

mjhuber:
I added support for ignoring audit controls.  Adding check ids to the `skipped_checks` variable will cause them to not affect the benchmark scoring.


sudermanjr (approved):
Looks pretty awesome to me.  Have you tried this in any clusters yet?

sudermanjr:
One possible enhancement for the future: 

When writing charts with cronjobs I have sometimes added a `ci/test-values.yaml` that forces the cronjob to run right away.

In addition to that, I have made a hook that creates a one-time job from the cronjob.  This allows the e2e testing to actually prove that the job works, not just installing the cron.

For example:

https://github.com/reactiveops/charts/blob/master/incubator/opa/templates/ca-updater-hook.yaml

mjhuber:
> One possible enhancement for the future:
> 
> When writing charts with cronjobs I have sometimes added a `ci/test-values.yaml` that forces the cronjob to run right away.
> 
> In addition to that, I have made a hook that creates a one-time job from the cronjob. This allows the e2e testing to actually prove that the job works, not just installing the cron.
> 
> For example:
> 
> https://github.com/reactiveops/charts/blob/master/incubator/opa/templates/ca-updater-hook.yaml

that's a great idea!  I'll have to think about how to let it work without datadog creds.

mjhuber:
> Looks pretty awesome to me. Have you tried this in any clusters yet?

thanks!  yes, i've been using it in a digital ocean & kops/aws cluster.

PR #71, opened 2019-04-29
Update ecr-cleanup to new version

mjhuber:
I bumped ecr-cleanup to version `0.1.6` and added the option `--keep-filters` flag.

### DIFF
--- a/stable/ecr-cleanup/templates/deployment.yaml
+++ b/stable/ecr-cleanup/templates/deployment.yaml
@@ -34,8 +34,8 @@ spec:
             - --v={{ .Values.controller.verbosity }}
             - --interval={{ .Values.controller.interval }}
             - --registry-id={{ .Values.controller.registryID }}
-{{- if .Values.controller.image_whitelist }}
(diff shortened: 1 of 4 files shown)


stable/ecr-cleanup/templates/deployment.yaml
```diff
@@ -34,6 +34,9 @@ spec:
             - --v={{ .Values.controller.verbosity }}
             - --interval={{ .Values.controller.interval }}
             - --registry-id={{ .Values.controller.registryID }}
+{{- if .Values.controller.image_whitelist }}
```
davekonopka: I might be overthinking it a bit but `whitelist` feels misleading since it's a list of things that shouldn't be cleaned up. What would you think about renaming it to `keep_filters` to match the underlying argument?

mjhuber: 👍 sounds good to me.  Updated to `keep_filters`.

sudermanjr (approved):
LGTM

PR #74, opened 2019-05-01
Get kind from releases and accept external PRs for testing

### LINKED ISSUE #64: Change CI to stop building kind
Kind is releasing binaries now, so we can stop building them.

### LINKED ISSUE #73: e2e tests cannot check out forked branches
This breaks external PR e2e testing.

sudermanjr:
Fixes #73
Fixes #64

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,23 +12,6 @@ references:
         apk --update add bash jq
 
 jobs:
-  # Because kind doesn't publish a proper binary, and we
-  # can't really install golang in alpine without pain,
-  # we build the kind binary from source and cache it.
-  build-kind:
-    docker:
-      - image: golang:1.12-stretch
-    steps:
-      - run:
-          name: Get and Build Kind
-          command: |
-            mkdir /binaries
-            go get sigs.k8s.io/kind
-            cp /go/bin/kind /binaries/kind
-      - save_cache:
-          key: v3-binaries
-          paths:
-            - /binaries
   lint-scripts:
     docker:
       - image: koalaman/shellcheck-alpine:v0.6.0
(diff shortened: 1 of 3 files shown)


sudermanjr:
This has been tested with a remote fork and worked.

.circleci/config.yml
```diff
@@ -69,10 +52,14 @@ jobs:
           name: Setup e2e Environment
           command: |
             cp /binaries/kind /usr/local/bin/kind
```
davekonopka: Should this be removed now?

davekonopka (approved):
One question inline, otherwise 👍

.circleci/config.yml
```diff
@@ -69,10 +52,14 @@ jobs:
           name: Setup e2e Environment
```
endzyme, line 49: Line 43 on the new file (restore_cache:) -- is that needed anymore since the save_cache above is removed?

.circleci/config.yml
sudermanjr: Yes, thank you.

.circleci/config.yml
sudermanjr: Yes, thank you.

.circleci/config.yml
```diff
@@ -69,10 +52,14 @@ jobs:
...
           command: |
             cp /binaries/kind /usr/local/bin/kind
+            curl -LO https://github.com/kubernetes-sigs/kind/releases/download/0.2.1/kind-darwin-amd64
+            chmod +x kind-darwin-amd64
+            mv kind-darwin-amd64 /usr/local/bin
```
endzyme: Is this running in a linux container with a Mac built binary?

sudermanjr: It does look that way.  I'm not sure how this is working right now.  🤔

endzyme: Also this will make a file called `/usr/local/bin/kind-darwin-amd64` but the script below references just `kind` as the binary name. I think maybe your restore cache may have been leading to a false positive in the testing?

sudermanjr: I think you're right.  I have fixed both of those in my latest iteration.

endzyme (approved):
lookin' sharp! nice!

PR #77, opened 2019-05-09
Improvements to papertrail liveness probe

robscott:
This liveness probe appears to be behaving at least somewhat better than the previous one.


sudermanjr (approved):
Sounds good to me.

PR #80, opened 2019-05-20
[incubator/fluentd] Papertrail improvements

coreypobrien:
- Fix logic for livenessprobe
- Fix configuration to use configured hostname/port
- Make sure to pass in hostname of node for non-kubernetes logs
- Use the actual hostname in combination with the clustername for non-kubernetes logs
- All records needs a hostname not just audit logs

### DIFF
--- a/incubator/fluentd/templates/daemonset.yaml
+++ b/incubator/fluentd/templates/daemonset.yaml
@@ -49,7 +49,7 @@ spec:
               fi; if [ -z "$(ls -A /var/log/fluentd-buffers-papertrail)" ]; then
                 echo "/var/log/fluentd-buffers-papertrail is empty. There are no buffers...I think that is good?";
                 exit 0;
-              fi; if [[ -z $(find /var/log/fluentd-buffers-papertrail -mmin -${LIVENESS_THRESHOLD_MINUTES}) ]]; then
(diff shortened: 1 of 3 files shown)


davekonopka (approved):
LGTM

incubator/fluentd/templates/daemonset.yaml
```diff
@@ -49,7 +49,7 @@ spec:
               fi; if [ -z "$(ls -A /var/log/fluentd-buffers-papertrail)" ]; then
                 echo "/var/log/fluentd-buffers-papertrail is empty. There are no buffers...I think that is good?";
                 exit 0;
-              fi; if [[ -z $(find /var/log/fluentd-buffers-papertrail -mmin -${LIVENESS_THRESHOLD_MINUTES}) ]]; then
+              fi; if [ -n "$(find /var/log/fluentd-buffers-papertrail -mmin ${LIVENESS_THRESHOLD_MINUTES} -name *.log)" ]; then
```
davekonopka, line 52: Is the flip from `-z` to `-n` intentional?

coreypobrien: Yep. The intent is to be `true` if the results from a search for a file that was last modified 15 minutes ago returns a string with a length > 0.

PR #86, opened 2019-05-24
[incubator/fluentd] Limit program names for papertrail to 32 characters

coreypobrien:
Papertrail requires `program` to be <= 32 chars: https://github.com/papertrail/remote_syslog_logger#default-program-name


dosullivan (approved):
Looks good to me, though I haven't had a chance to test it. Seems like it should be a safe change.

PR #92, opened 2019-06-14
update polaris to use minor versions

rbren:


sudermanjr (commented):
As long as Polaris is truly committed to non-breaking minor version changes, then this seems fine, but if it isn't, then I would really prefer to pin minor rather than major.

sudermanjr (approved):

PR #96, opened 2019-06-27
add audit configuration for Polaris

rbren:


stable/polaris/templates/audit.rbac.yaml
```diff
@@ -0,0 +1,95 @@
...
+  {{- end }}
+  labels:
+    app: polaris
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: polaris-cleanup
+subjects:
+  - kind: ServiceAccount
+    name: polaris-audit
```
davekonopka: I think the name here should match the others: `{{ include "polaris.fullname" . }}-audit`

stable/polaris/templates/audit.job.yaml
```diff
@@ -0,0 +1,70 @@
...
+  template:
+    spec:
+      serviceAccountName: {{ include "polaris.fullname" . }}-audit
+      restartPolicy: Never
+      containers:
+      - command:
+        - polaris
+        - --audit
+        - --output-url
+        - {{ .Values.audit.outputURL }}
```
davekonopka: Can you add some info to the values file or to the readme describing what expected `outputURL` values look like?

davekonopka (approved):
A few minor comments in line. Otherwise 👍🏻

stable/polaris/templates/audit.job.yaml
endzyme: Agreed - this breaks installing by just setting `audit.enable=true` with the following error: 

```
UPGRADE FAILED
Error: error validating "": error validating data: ValidationError(Job.spec.template.spec.containers[0].command): unknown object type "nil" in Job.spec.template.spec.containers[0].command[3]
Error: UPGRADE FAILED: error validating "": error validating data: ValidationError(Job.spec.template.spec.containers[0].command): unknown object type "nil" in Job.spec.template.spec.containers[0].command[3]
```

endzyme: It may be a better error to add a `required` function if enabled is true. Example

```gotemplate
        - --output-url
        - {{ required "Must set audit.outputURL in values if you enable the audit job." .Values.audit.outputURL }}
        - --output-file
```

stable/polaris/templates/audit.job.yaml
```diff
@@ -0,0 +1,70 @@
...
+            drop:
+              - ALL
+        volumeMounts:
+        - name: config
+          mountPath: /opt/app/config.yaml
+          subPath: config.yaml
+          readOnly: true
+        - name: results
+          mountPath: /tmp/results
+      - name: cleanup
```
kimschles, line 55: Nick and I played around with this, and we found that this deletes the entire namespace. Is this the intended behavior?

stable/polaris/templates/audit.job.yaml
rbren: Thanks Nick! I didn't know about the `required` function - works much better now.

stable/polaris/templates/audit.job.yaml
rbren: Yeah that's intended, but perhaps a bit scary. I added a `cleanup` variable, which defaults to false.

rbren:
Thanks all for the feedback! I think I addressed everything.

endzyme (approved):
thanks again for syncing up on video!

PR #100, opened 2019-07-11
feat(stable:rbac-manager): Upgrade rbac-manager to latest version

oxyno-zeta:
Upgrade rbac-manager to latest version and some readability on README and values.

### DIFF
--- a/stable/rbac-manager/Chart.yaml
+++ b/stable/rbac-manager/Chart.yaml
@@ -1,6 +1,6 @@
 name: rbac-manager
-version: 1.1.2
-appVersion: 0.6.1
+version: 1.2.0
+appVersion: 0.7.0
 description: A Kubernetes operator that simplifies the management of Role Bindings and Service Accounts.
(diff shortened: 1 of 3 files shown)


stable/rbac-manager/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 name: rbac-manager
-version: 1.1.2
-appVersion: 0.6.1
+version: 1.1.3
```
sudermanjr: Since this is a minor version change to the underlying app, can you please make this a minor version change in the chart?

sudermanjr (changes requested):
Since this is a minor version change to the underlying app, can you please make this a minor version change in the chart? I'm thiking 1.2.0 would be appropriate here.

oxyno-zeta:
Of course I can :) . Not a problem.
This is done.

sudermanjr (approved):
Thanks!

PR #102, opened 2019-07-18
fluentd chart to API v1

crmejia:
The API has changed. See: https://docs.fluentd.org/quickstart/update-from-v0.12
Might as well bite the bullet and migrated the chart to the new version. Breaking change as I renamed some values.

### DIFF
--- a/incubator/fluentd/README.md
+++ b/incubator/fluentd/README.md
@@ -17,12 +17,12 @@ Switching backends requires using a matching image as well as some backend-speci
 | `elasticsearch.port` | `443` |
 | `elasticsearch.scheme` | `https` |
 | `elasticsearch.extraArgs` | [] |
-| `elasticsearch.buffer_chunk_limit` | `5m` |
-| `elasticsearch.buffer_queue_limit` | `100` |
-| `elasticsearch.buffer_queue_full_action` | `exception` |
-| `elasticsearch.request_timeout` | `20s` |
+| `elasticsearch.chunk_limit_size` | `5m` |
+| `elasticsearch.chunk_limit_records` | `100` |
+| `elasticsearch.overflow_action` | `throw_exception` |
+| `elasticsearch.retry_timeout` | `20s` |
 | `elasticsearch.flush_interval` | `5s` |
-| `elasticsearch.num_threads` | `4` |
+| `elasticsearch.flush_thread_count` | `4` |
 
 
 ### Papertrail
@@ -30,7 +30,7 @@ Switching backends requires using a matching image as well as some backend-speci
 | Parameter | Default |
 | --------- | ------- |
 | `image.repository` | `fluent/fluentd-kubernetes-daemonset` |
-| `image.tag` | `v0.12.43-debian-papertrail` |
+| `image.tag` | `v1.4.2-debian-papertrail-1.1` |
 | `papertrail.enabled` | `true` |
 | `papertrail.host` | `logs3.papertrailapp.com` |
 | `papertrail.port` | `12785` |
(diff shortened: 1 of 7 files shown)


incubator/fluentd/values.yaml
```diff
@@ -33,13 +32,16 @@ papertrail:
...
-  extraArgs: []
 
 
 #####################
 ## Shared Settings ##
 #####################
 
+
+#plugin command extra args
+pluginExtraArgs: []
```
crmejia, line 44: I moved this one outside of Elastic as it can also be used by papertrail.

incubator/fluentd/files/papertrail.conf
```diff
@@ -29,9 +29,11 @@
 <match **>
   @type papertrail
   @id out_papertrail
-  num_threads 4
-  buffer_type    file
-  buffer_path    /var/log/fluentd-buffers-papertrail/fluentd-buffer
+  <buffer>
+    @type    file
+    path    /var/log/fluentd-buffers-papertrail/fluentd-buffer
+    flush_thread_count 4
```
dosullivan: Is there a reason for `flush_thread_count` to be hard-coded here, but not in the elasticsearch.conf file? Or could this have a templated value set from `values.yaml` as well?

incubator/fluentd/README.md
```diff
@@ -16,21 +16,20 @@ Switching backends requires using a matching image as well as some backend-speci
...
+| `elasticsearch.flush_thread_count` | `4` |
 
 
 ### Papertrail
 
 | Parameter | Default |
 | --------- | ------- |
 | `image.repository` | `fluent/fluentd-kubernetes-daemonset` |
-| `image.tag` | `v0.12.43-debian-papertrail` |
+| `image.tag` | `v1.4.2-debian-papertrail-1.1` |
```
dosullivan, line 32: We should update the image.tag value under the list of elasticsearch parameters too right?

incubator/fluentd/values.yaml
```diff
@@ -33,13 +32,16 @@ papertrail:
   enabled: false
```
dosullivan, line 32: Can the image.tag be updated here too (line 29), even though this is commented out? Would be good for consistency, to imply that the new version is the new minimum everywhere.

incubator/fluentd/files/elasticsearch.conf
```diff
@@ -5,18 +5,18 @@
...
+    @type memory
+    chunk_limit_size {{ .Values.elasticsearch.chunk_limit_size }}
+    chunk_limit_records {{ .Values.elasticsearch.chunk_limit_records }}
+    flush_interval {{ .Values.elasticsearch.flush_interval }}
+    flush_thread_count {{ .Values.elasticsearch.flush_thread_count }}
+    retry_forever
+    retry_wait 1s
+    retry_max_interval 60s
+    retry_timeout {{ .Values.elasticsearch.retry_timeout }}
+    overflow_action {{ .Values.elasticsearch.buffer_queue_full_action }}
```
dosullivan: I think this needs to be updated to `.Values.elasticsearch.elasticsearch.overflow_action` to line up with the values.yaml now, and with the updated parameter docs above

dosullivan (changes requested):
I think there's a couple parameters that need to be lined up still (see individual comments) but on whole this looks good to me!

incubator/fluentd/files/elasticsearch.conf
crmejia: that's a good catch!

PR #104, opened 2019-07-19
[stable/rbac-manager] Sync liveness/readiness and securitycontext

coreypobrien:
All these changes are from https://github.com/FairwindsOps/rbac-manager/blob/master/deploy/all.yaml


sudermanjr (approved):
Thanks!

PR #106, opened 2019-07-24
Adjustments for 0.4 (polaris PR 166)

endzyme:
- Added version bump to 0.4
- Added new RBAC rules


endzyme:
@kimschles @bobby-brennan this is ready for review now that the release is done and the tests are passing :D

rbren (approved):
I think the README needs an update with the new image tags, but otherwise LGTM

PR #110, opened 2019-07-30
support nodeselector and toleration

bambash:
Added support for nodeSelector and tolerations within the dashboard and webhook deployments.
Updated webhook secret name to be templated.


rbren (approved):
Looks good! Can you just add a note in the readme before merging?

PR #113, opened 2019-08-05
fix: home and source URLs in the Chart.yaml

tklovett:
This is intended to fix the links shown for the chart on https://hub.helm.sh/charts/reactiveops-stable/rbac-manager


sudermanjr (approved):
Thanks for the contribution!

PR #114, opened 2019-08-08
Adding goldilocks chart

sudermanjr:


dosullivan (approved):
This all looks good to me, templated correctly on my machine. Would it be appropriate to include a note in the goldilocks readme that it has dependency on the VPA?

sudermanjr:
e2e tests on this one are going to be difficult due to the depency on VPA.  I will do some thinking on it.

sudermanjr:
@bambash I'm going to need your review on this one until the CODEOWNERS file is merged.

PR #115, opened 2019-08-09
[stable/aws-iam-authenticator]: Add `resources` values

ivanfetch:
This parameterizes the resource requests and limits for the aws-iam-authenticator pods.


sudermanjr (approved):
LGTM. Thanks.

PR #117, opened 2019-08-14
Fixing permissions for goldilocks dashboard

sudermanjr:
The dashboard didn't have permissions to view deployments, which is what allows it to see existing resource requests/limits as well as see the container names.  

This will give the dashboard those permissions


dosullivan (approved):
✅

mgoodhart5 (approved):

PR #120, opened 2019-08-19
WIP: Update ro-cert-manager chart for 0.8.1

lucasreed:


sudermanjr (changes requested):
Can you update the `requirements.yaml`  file to install the 0.8.1 version of cert-manager when cert-manager install is enabled?

Also, can we add a `ci/test-values.yaml` that will install cert-manager and test it with the new version of cert-manager with a solver-based clusterissuer setup?

lucasreed:
@sudermanjr I did start to do that but circleci would fail when I changed the requirements.yaml to pull from jetstack instead of the default helm repo (which is how it currently runs). I kinda forgot about that after I changed it back, thanks for the reminder. I'll work on this cert-manager stuff tomorrow morning.

ivanfetch (approved):
Thanks for making this update in support of future 1.X cert-manager versions! While I think comments calling out which things are deprecated would be a nice thing to have added to the chart values, it doesn't block this pull request.

endzyme (approved):
Non-blocking request: Can we add a changelog with some context here on migration path? I know there's been discussion on if we need to purge delete and I wanted to capture some of the version changes and intent somewhere. 

Otherwise looks great!!!

lucasreed:
Changing this to WIP. I need to do some reworking on here because theoretically it can support multiple DNS providers and choose which one to use based on the `selector` field of a given dns `solver`.

Do we need this functionality? Do we have customers that utilize two different DNS providers in a single kube cluster?

@sudermanjr  @ivanfetch @endzyme

sudermanjr (approved):

PR #123, opened 2019-09-09
adding letsencrypt-setup and e2e pre-test hook

lucasreed:
A lot of changes, including naming conventions, for setting up ClusterIssuers alongside [cert-manager by jetstack](https://github.com/jetstack/cert-manager).

Name changes:
ro-cert-manager -> letsencrypt-setup
staging -> selfsigned
production -> primary

Open for suggestions on any names that would be better.

Another change is that I added `cloudflare` as a first class citizen in the dns solvers section. We have a customer in pod-2 that uses this. We could also add Azure DNS when necessary. I don't have a good way to test that currently.

I tried my best to document the various DNS solvers and how to configure them in the README.md. This includes the new `selector` object that is present in every solver config.

I have also added a hook into the e2e tests to allow a chart author to create `ci/pre-test-script.sh` within their chart to have some arbitrary things (in this case, installing remote CRD's) run before the test. Open to other ideas here in case this is frowned upon. The script is run from within the `executor` container during the e2e-test phase in circleci.


.circleci/config.yml
```diff
@@ -23,6 +23,12 @@ jobs:
             for script in $(ls scripts/*.sh); do
                 shellcheck -x $script
             done
+      - run:
+          name: lint-pre-scripts
+          command: |
+            for script in $(find incubator stable -type f -name "*.sh"); do
```
sudermanjr: there may be other scripts in these folders for various reasons that are part of the helm chart.  I don't think we want to lint those, so maybe a more specific match would be good here?

incubator/cert-manager-setup/ci/test-values.yaml
```diff
@@ -0,0 +1,10 @@
+clusterIssuers:
```
sudermanjr: AFAICT this does test the DNS solver configuration at all.  Can this be expanded to do that?

scripts/e2e-test.sh
```diff
@@ -47,6 +47,7 @@ setup_executor () {
     echo "$port"
     docker inspect "$containerName" | jq .
     docker exec "$EXEC_CONTAINER_NAME" sh -c "sed -i 's/https:\/\/localhost:[0-9]\+/https:\/\/localhost:6443/g' /.kube/config"
+    docker exec "$EXEC_CONTAINER_NAME" sh -c "sed -i 's/https:\/\/127.0.0.1:[0-9]\+/https:\/\/127.0.0.1:6443/g' /.kube/config"
```
sudermanjr, line 50: Can you explain the purpose of adding this (or why it was necessary)?

scripts/pre-test-script-runner.sh
```diff
@@ -0,0 +1,36 @@
...
+set -o errexit
+set -o nounset
+set -x
+
+SCRIPT_PATH="$(cd "$(dirname "$0")" || return ; pwd -P)"
+CI_DIR="ci"
+SCRIPT_NAME="pre-test-script.sh"
+SCRIPT="$CI_DIR/$SCRIPT_NAME"
+
+for chart in "$SCRIPT_PATH"/../incubator/* ; do
```
sudermanjr: I wonder if it is possible to get something from the lint script that tells us which chart is being tested rather than just looking for any pre-script. Won't this run all of them if it finds them?

sudermanjr (changes requested):

scripts/e2e-test.sh
lucasreed: I had to add this for my local testing. Docker appeared to use the IP instead of `localhost` for me. I figured it couldn't hurt to have it here, but I can remove it if we'd rather keep it minimal.

scripts/pre-test-script-runner.sh
lucasreed: Yes, currently it will run any of them if found. I will look into this.

.circleci/config.yml
lucasreed: Changed to only check pre-test-script.sh in any chart's ci folder.

incubator/cert-manager-setup/ci/test-values.yaml
lucasreed: Added dns solver

scripts/pre-test-script-runner.sh
lucasreed: I implemented this without any output from the linting tool because I couldn't find any docs for the `ct lint` command that would provide the info we need.

lucasreed:
Adjusted the title to include [DO NOT MERGE YET] so we can discuss if a name change makes sense.

scripts/e2e-test.sh
sudermanjr: Nope, seems fine to keep it. Just wanted to check.

sudermanjr (approved):
Looks amazing.  Thanks! @bambash  can you also take a look since this modifies the e2e process slightly?

bambash (approved):
awesome work, looks great

PR #124, opened 2019-09-10
[incubator/helm-release-pruner] Use image and improve cronjob

coreypobrien:
- Use a dedicated image to be able to use a read-only filesystem
- Run as non-root
- Add a starting deadline to prevent schedule failure
- Add a concurrency policy
- Remove priv escalation and sys caps
- Add resource requests


incubator/helm-release-pruner/templates/configmap.yml
```diff
@@ -6,110 +6,6 @@ data:
   start-up.sh: |-
     #!/bin/sh
 
-    apk add --update bash coreutils
-
 {{- range .Values.pruneProfiles }}
     /usr/local/bin/prune-releases.sh --older-than="{{ .olderThan }}" --helm-release-filter="{{ .helmReleaseFilter }}" --namespace-filter="{{ .namespaceFilter }}"{{ if $.Values.job.dryRun }} --dry-run{{ end }}
```
sudermanjr, line 10: In general I think we should make this script the entrypoint for the container and just skip start script if we're going to do it with a custom image.

Would be nice to make all the parameters ENV for the new container. Not blocking for this PR

sudermanjr (approved):
LGTM for now.  I'll plan on some more updates going forward.

mjhuber:
it might make sense to keep the `appVersion` consistent with the helm version because the output has varied slightly between versions before

sudermanjr:
> it might make sense to keep the `appVersion` consistent with the helm version because the output has varied slightly between versions before

I think it's probably better to stick to convention and keep the appVersion consistent with the docker image that we are installing.  That being said, maybe we should update the app version of the docker image to be higher than it is now to indicate the changes.

mjhuber (approved):

incubator/helm-release-pruner/templates/configmap.yml
coreypobrien: Yeah that definitely gets complicated because of the chart having a variable length loop. The loop params either have to explicitly generate the loop steps as they do now via helm or they would have to be passed into the container which could get messy with 3 parameters per iteration (comma separated values?).

PR #127, opened 2019-09-12
Adding Fairwinds Custom Metrics chart

sudermanjr:


bambash (approved):
:ok_hand: looks good, chart installs cleanly

PR #136, opened 2019-10-03
updating app version for letsencrypt-setup; adding me as an owner

lucasreed:


incubator/letsencrypt-setup/Chart.yaml
```diff
@@ -1,8 +1,9 @@
 name: letsencrypt-setup
 apiVersion: v1
-version: v1.0.1
-appVersion: v0.8.1
+version: v1.0.2
```
sudermanjr: Since this is a minor version app change, I would usually recommend a minor version chart update.  Up to you since I know LE doesn't change much from 0.8 to 0.9

lucasreed: Oh good point. I'll do that.

sudermanjr (approved):

PR #140, opened 2019-10-10
set ingressClass in default values

mjhuber:
I set the `ingressClass` to `nginx` in values.yml.  When this is not set and the http solver is enabled the ingress object in the issuer gets set to `null`, causing cert-manager to fail on every http01 validation:
```
cert-manager  Error presenting challenge: challenge's 'solver' field is specified but no HTTP01 ingress config provided. Ensure solvers[].http01.ingress is specified on your issuer resource
```
Setting this to `nginx` by default will allow the http01 solver to work out of the box with the default ingress class setting for nginx-ingress.


bambash (approved):
looks good! :tada:

PR #141, opened 2019-10-11
Add chart for Fairwinds Insights

rbren:


rbren:
I'd rather this chart _not_ build unless certain flags are specified (specifically, `organization`, `cluster`, and `base64token`). Is there a way to fix the tests without adding defaults in `values.yaml`?

bambash (commented):
this looks great! 

I'm not entirely sure about testing, but you may be able to fix the CI process by following this pattern: https://github.com/FairwindsOps/charts/tree/master/incubator/letsencrypt-setup/ci

I noticed a few things that are statically defined which would be better served via templating.
* resources in `_job.spec` - it would be nice to override these via `values.yaml`
* non-templated `metadata.name` in all `rbac.yaml`
* non-templated `metadata.name` and `metadata.labels` in `serviceaccount.yaml`
* non-templated `metadata.name` in `token-secret.yaml`

It looks like a goldilocks chart tarball is pushed to `insights-agent/charts`, not sure if that is needed

sudermanjr:
> certain flags are specified (specifically, `organization`, `cluster`, and `base64token`). Is there a way to fix the tests without adding defaults in `values.yaml`?

Whatever you specify in `ci/*-values.yaml` will be used during e2e testing. So you can tweak the tests that way.

scripts/ct.yaml
```diff
@@ -5,4 +5,5 @@ chart-dirs:
   - incubator
 chart-repos:
   - "jetstack=https://charts.jetstack.io"
+  - "fairwinds-incubator=https://charts.fairwinds.com/incubator"
```
sudermanjr: You're going to want to add stable here too. Goldilocks is moving to stable

stable/insights-agent/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+appVersion: 0.0.1
+description: A Helm chart to run the Fairwinds Insights agent
+name: insights-agent
+version: 0.4.0
```
sudermanjr, line 5: Curious why the chart is starting at 0.4.0 and the app at 0.0.1.  I would expect the opposite.

stable/insights-agent/Chart.yaml
```diff
@@ -0,0 +1,7 @@
...
+maintainers:
+  - name: bobby-brennan
```
sudermanjr: Are you going to be the only maintainer?

stable/insights-agent/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+The Fairwinds Insights Agent is a collection of reporting tools, which send data back
+to [Fairwinds Insights](https://insights.fairwinds.com). If you'd like access
+to Fairwinds Insights, contact `opensource@reactiveops.com`
+
+## Installation
+We recommend installing `insights-agent` in its own namespace.
+
+```
+helm repo add fairwinds https://charts.fairwinds.com/stable
```
sudermanjr: We generally try to stick to the `fairwinds-stable` convention due to our customers using incubator frequently.

stable/insights-agent/requirements.lock
```diff
@@ -0,0 +1,6 @@
+dependencies:
+- name: goldilocks
+  repository: https://charts.fairwinds.com/incubator
```
sudermanjr: Once #145 gets merged, you will want to change this. My guess is it will merge before this one.

stable/insights-agent/requirements.yaml
```diff
@@ -0,0 +1,8 @@
+dependencies:
+  - name: goldilocks
+    repository: https://charts.fairwinds.com/incubator
+    version: 0.2.0
+    condition: goldilocks.enabled
+    values:
```
sudermanjr: I didn't know you could do this.  Neat.

stable/insights-agent/templates/token-secret.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
```
sudermanjr, line 2: A lot of people (including Fairwinds) don't like create secrets using Helm due to how it stores those values in the cluster.  I would like to see this chart at least provide the option to specify a pre-existing secret name.

sudermanjr (changes requested):
I agree with _Nate on the desire to see more templating around Rbac names and such. Also, I left some comments inline.

scripts/ct.yaml
rbren: Done

stable/insights-agent/Chart.yaml
rbren: This is a weird chart, since it involves several different "apps" (polaris, kube-hunter, goldilocks, ...).

I made `appVersion` `0.0.12` which matches the current version of the Insights application (which, ironically, is _not_ the application installed by this agent. but it's where the data is going, so...)

stable/insights-agent/Chart.yaml
rbren: Added Barnabas

stable/insights-agent/README.md
rbren: Done

stable/insights-agent/requirements.lock
rbren: Done

stable/insights-agent/requirements.yaml
rbren: In fact, you cannot. :(

Moved to values.yaml

stable/insights-agent/templates/token-secret.yaml
rbren: Makes sense. Done!

rbren:
Think I addressed everything. Let me know if there's something I missed.

Thanks for all the feedback!

stable/insights-agent/Chart.yaml
sudermanjr: ¯\(◉‿◉)/¯

stable/insights-agent/requirements.yaml
sudermanjr: 😆

sudermanjr (commented):
Looks like the build failed during docker hub outage. I am re-running

sudermanjr (approved):
Looks good! I'm sure there are bugs that got missed (it's a big chart), but good to go for this PR assuming the build passes this time.

PR #146, opened 2019-10-14
Update CODEOWNERS

rbren:


sudermanjr:
Since Kim is no longer a CO of polaris, do you want to remove her as well?

rbren:
👍 good point

sudermanjr (approved):

PR #150, opened 2019-10-18
Disable both ES and papertrail fluentd plugins by default

dosullivan:
I think it's a slightly-more-user-friendly behavior to make the user specifically choose the plugin they want enabled by default (be it papertrail or elasticsearch).

### DIFF
--- a/incubator/fluentd/Chart.yaml
+++ b/incubator/fluentd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.12.43"
 description: Fluentd for multiple endpoints
 name: fluentd
-version: 4.0.0
+version: 4.0.1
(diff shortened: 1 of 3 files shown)


sudermanjr:
> I think it's a slightly-more-user-friendly behavior to make the user specifically choose the plugin they want enabled by default (be it papertrail or elasticsearch).

Agreed, however can you note in the README (maybe in the backends section or the top-level section), that by default no backend is enabled?

sudermanjr:
You will need to version bump this chart to get the linter to pass and run the e2e tests.

dosullivan:
Added comment and version-bumped.

coreypobrien (approved):

sudermanjr (approved):

PR #154, opened 2019-10-21
[Goldilocks] Fix logos for helm hub

sudermanjr:


stable/goldilocks/README.md
```diff
@@ -1,5 +1,5 @@
 <div align="center">
-<a href="https://github.com/FairwindsOps/goldilocks"><img src="logo.svg" height="150" alt="Goldilocks" style="padding-bottom: 20px" /></a>
+<a href="https://github.com/FairwindsOps/goldilocks"><img src="https://raw.githubusercontent.com/FairwindsOps/goldilocks/master/pkg/dashboard/assets/images/goldilocks.svg" height="150" alt="Goldilocks" style="padding-bottom: 20px" /></a>
 <br>
```
mgoodhart5, line 3: This image is broken when I view the file, just wanted to mention it in case that was an issue.

dosullivan (approved):

stable/goldilocks/README.md
dosullivan: I think that's just SVG being funky when viewed directly in the browser. The helm chart.yaml spec allows for svg to be used in the `icon` field though, so I think it should be fine.

PR #155, opened 2019-10-23
Add a basePath value which passes --base-path arg to the dashboard container

out-of-band:
If you enable an ingress with path on this chart install, for instance using `my.cluster.com` with a path `/goldilocks`, the dashboard will not work because resources are requested as e.g., `my.cluster.com/main.js` rather than `my.cluster.com/goldilocks/main.js`.

This just allows the user to set the `basePath` which then gets passed along in the dashboard container's command.


sudermanjr (changes requested):
@out-of-band This is a great addition, thanks!  Can you please add the new value to the table in the README?

sudermanjr (approved):

PR #159, opened 2019-11-04
Enable goldilocks in insights-agent

rbren:
Can break out the goldilocks change if you prefer Andy


stable/insights-agent/requirements.yaml
```diff
@@ -1,5 +1,5 @@
 dependencies:
   - name: goldilocks
     repository: https://charts.fairwinds.com/stable
-    version: 1.0.0
+    version: 1.1.1
```
sudermanjr, line 4: Seems like at least a minor version bump for the change in Goldilocks to v2 and enabling it by default.

sudermanjr:
You might have to separate these PRs just to get the linter to pass. The Goldilocks bump in insights can't pull the new version of goldilocks because it doesn't exist yet.

makoscafee (approved):

sudermanjr (changes requested):
Not sure why, but the e2e tests are not getting any output whatsoever on this chart:

```
Installing chart with values file 'stable/insights-agent/ci/test-values.yaml'...

Too long with no output (exceeded 10m0s)
```

sudermanjr (approved):

PR #160, opened 2019-11-04
fix quoted flags for goldilocks

rbren:


rbren:
Anything more I can do testing-wise?

sudermanjr (approved):
This looks great.  If you wanted to add your other new flags to the e2e testing I wouldn't complain, but this is great for now.

PR #163, opened 2019-11-06
Update polaris

rbren:


stable/polaris/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 description: Validation of best practices in your Kubernetes clusters
 name: polaris
-version: "0.6.1"
-appVersion: "0.4"
+version: "0.6.2"
+appVersion: "0.5"
```
sudermanjr, line 5: I usually prefer to link the minor version changes, i.e. if polaris goes up a minor, then the chart should go up a minor, but that's entirely up to you all as the owners of the chart.

rbren: SGTM!

sudermanjr (approved):

PR #166, opened 2019-11-07
randomize minute for reports

rbren:


makoscafee (commented):
Overall looks good!. Just curious why randomize the time ?

rbren:
Good question! I added a note to the README as well.

Randomizing the time helps both the client (because they won't suddenly have four reporting tools taking up resources at the same time) and the server at `insights.fairwinds.com` (because we won't get an onslaught of reports at the top of every hour).

makoscafee (approved):

PR #167, opened 2019-11-08
run uploader using nonRootUser

makoscafee:


rbren (approved):
Great job tracking this down!

PR #168, opened 2019-11-12
set runAsUser to fix insights agent uploader

rbren:


rbren:
Pushing this through, as the chart is currently broken

PR #170, opened 2019-11-12
update goldilocks resources based on goldilocks recommendations

rbren:


makoscafee (approved):
Looks good... but will these be changing based on the cluster/node memory size?

rbren:
I don't think so, no. @sudermanjr any thoughts on this?

I suppose they could change if you have a lot of deployments you're using GL for - i.e. if GL has to do a lot more work. But I don't think that'd be too excessive.

I'm noticing that VPA's `upperBound` is extraordinarily high. I may add a last-minute change to use `uncappedTarget` instead.

sudermanjr:
> I don't think so, no. @sudermanjr any thoughts on this?
> 
> I suppose they could change if you have a lot of deployments you're using GL for - i.e. if GL has to do a lot more work. But I don't think that'd be too excessive.

I think it would have to a _lot_ of deployments to blow up too bad.  The list is held in memory, so it could happen, but they are pretty small objects.

I can test it out if it's something you are worried about.

rbren:
No worries - just wanted to make sure the requests/limits we're setting here are broadly applicable, rather than specific to the KIND cluster I tested on.

PR #172, opened 2019-11-13
Bb/k8s node termination handler

bbensky:
Adding a chart for k8s-node-termination-handler, so GKE clusters can incorporate graceful node draining on preemptible nodes when GCP terminates them automatically.

This is the repo for k8s-node-termination-handler: [https://github.com/GoogleCloudPlatform/k8s-node-termination-handler]


incubator/k8s-node-termination-handler/templates/daemonset.yaml
```diff
@@ -0,0 +1,72 @@
+# Copyright 2018 Google Inc. All Rights Reserved.
```
sudermanjr: Is this copy/pasta from somewhere? I don't think we should be licensing our stuff under google....

bbensky: I removed those blocks from daemonset and rbac yamls

sudermanjr (approved):
LGTM. Thanks!

incubator/gke-node-termination-handler/README.md
```diff
@@ -0,0 +1,19 @@
...
+We recommend installing gke-node-termination-handler in its own namespace.
+
+## Configuration
+Parameter | Description | Default
+--------- | ----------- | -------
+`image.repository` | Image repo | k8s.gcr.io/gke-node-termination-handler@sha256
+`image.tag` | Image tag | aca12d17b222dfed755e28a44d92721e477915fb73211d0a0f8925a1fa847cca
+`image.pullPolicy` | Image pull policy | IfNotPresent
+`daemonset.updateStrategy.type` | Daemonset update strategy type | RollingUpdate
+`resources.limits.cpu` | Daemonset CPU limit | 150m
```
sudermanjr, line 16: For future improvement, I think it's easier to use an entire resources block instead of each value independently.  If you do a `helm create <chart name>` you'll see the pattern I am referring to.

PR #173, opened 2019-11-13
Changing over to just install the recommender, CRD, and RBAC. Add an upgrade path

sudermanjr:


rbren:
Why no more `upgradeVPA` flag? Guessing that'd just be identical to running `installVPA` a second time?

Can we call it something other than `uninstallVPA`? Maybe `reinstallVPA`, or `uninstallOldVPA`?

My plan is to add those flags to the default insights-agent install command for folks running appropriate versions of the chart - it'd look a little broken to have both `--set installVPA=true --set uninstallVPA=true`

sudermanjr:
> Why no more `upgradeVPA` flag? Guessing that'd just be identical to running `installVPA` a second time?
> 
> Can we call it something other than `uninstallVPA`? Maybe `reinstallVPA`, or `uninstallOldVPA`?
> 
> My plan is to add those flags to the default insights-agent install command for folks running appropriate versions of the chart - it'd look a little broken to have both `--set installVPA=true --set uninstallVPA=true`

Yeah, so the upgrade is now identical due to the `kubectl apply` being in there. That's why it was removed.

We _could_ call it reinstallVPA, but it doesn't really do that, it just uninstalls.  I suppose we _could_ just make it uninstall and install in a single go, but I wanted the option to make it a post-delete hook as well just in case. 

How about `preUninstallVPA` or something like that?

sudermanjr:
I added a `reinstallVPA` flag that combines the two in order to make the command more palatable

stable/goldilocks/templates/vpa-uninstall-hook.yaml
```diff
@@ -1,65 +1,65 @@
-{{- if .Values.upgradeVPA }}
+{{- if or (eq .Values.uninstallVPA true) (eq .Values.reinstallVPA true) }}
```
rbren: Maybe simplify to `if or .Values.uninstallVPA .Values.resinstallVPA`

rbren (approved):

stable/goldilocks/templates/vpa-uninstall-hook.yaml
sudermanjr: Oh sure, why not.  I was being lazy and copying from somewhere ;-)

PR #174, opened 2019-11-13
trying out labeler

sudermanjr:


bambash:
What is the context on this bad boy?

sudermanjr:
> What is the context on this bad boy?

I was testing to see if github actions could automatically label PRs based on what folder they came from.

PR #175, opened 2019-11-14
Added workload report for insights

makoscafee:


rbren:
Looks good! Just two things:
* You'll need to bump the chart version in Chart.yaml to get the tests to pass
* We should add `workloads` options to the README

makoscafee:
I have also updated the app version since we have `0.3.0` released

stable/insights-agent/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 description: A Helm chart to run the Fairwinds Insights agent
 name: insights-agent
-version: 0.7.6
-appVersion: 0.2.0
+version: 0.7.7
+appVersion: 0.3.0
```
rbren: This will actually go out with app version `0.4.0`

rbren (approved):

PR #178, opened 2019-11-15
[stable/insights] Updating goldilocks to v2.0.1 in insights

sudermanjr:


stable/insights-agent/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to run the Fairwinds Insights agent
 name: insights-agent
-version: 0.7.6
+version: 0.8.0
```
makoscafee, line 4: New to how this version thing works. Reason to skip to `0.8.0` ?

makoscafee (commented):
LGTM just a question on versioning. Otherwise looks good.

rbren:
Please **do not merge** this yet. The website will always instruct users to install the latest helm chart, so if the instructions need to change, we'll need to wait for a release there.

Thanks for putting this together Andy! Is there any reason not to set `goldilocks.installVPA=true` as a default in values.yaml? I believe it's idempotent now, and I think it will still be contingent on `goldilocks.enabled=true` since the dependency has that condition.

We currently have instructions on the Insights website to `--set goldilocks.installVPA=true` if you're installing for the first time. My plan is to only show `--set goldilocks.reinstallVPA=true` if the current chart version is < 0.8 - otherwise we won't instruct them to set any goldilocks-related flags.

How does that sound?

sudermanjr:
>Is there any reason not to set goldilocks.installVPA=true as a default in values.yaml?

I am hesitant to do so in the event that people are managing their own installation of VPA, however I might be convinced that this is a minor risk.

> My plan is to only show --set goldilocks.reinstallVPA=true if the current chart version is < 0.8

This is a good idea.

stable/insights-agent/Chart.yaml
sudermanjr: Since the major version of the goldilocks chart has gone from 1-2, I felt that this would be a potential significant change in insights and bumped the minor version instead of patch.

rbren:
Let's default to `installVPA=true`, and I'll add a note to the install instructions about setting to `false` if you've got VPA installed already.

rbren:
LGTM. Holding off on approval until we release an Insights update Wednesday AM, which will update chart install instructions.

rbren:
Pushing the release back to Monday.

Going to update the `helm install` instructions to pin the minor version, so we won't have to hesitate to create a new minor release of the helm chart.

rbren (approved):

rbren:
Getting this in ahead of today's release

PR #189, opened 2019-12-03
updating letsencrypt-setup chart version along with new apiVersion for use with cert-manager v0.11.0+

lucasreed:


sudermanjr:
What's the process for upgrading this chart? Is it expected that we will update the resources as part of the upgrade to v0.11+ and then this chart will run smoothly?

sudermanjr:
What's the process for upgrading this chart? Is it expected that we will update the resources as part of the upgrade to v0.11+ and then this chart will run smoothly?

lucasreed:
Yes, during the upgrade of cert-manager we take a backup of all the custom resources that cert-manager owns (including ClusterIssuers) and manually update the apiVersion and re-post them at the end. In fact, as I typed that out I think I ought to test that part too. I'll do that.

lucasreed:
If recreating the ClusterIssuers during the cert-manager upgrade using backed-up resource definitions breaks helm, it would be fine to leave ClusterIssuers out of the backup and run this new chart after cert-manager is at v0.11.X

sudermanjr (approved):

sudermanjr (changes requested):
Can we add a section to the readme that just mentions the upgrade process? I would be worried about people just bumping the version.

Maybe we should consider a major version bump on the chart for this to indicate that?

incubator/letsencrypt-setup/README.md
```diff
@@ -2,6 +2,12 @@
 
 Installs the necessary additional items that the cert-manager chart lacks.
 
+## Upgrading to 2.0.0 Chart Version
+
+There are breaking changes with 2.0.0 chart version in that it absolutely requires a version of cert-manager v0.11.0 or higher to be running in the cluster.
+
+Please follow upgrade instructions [located here](https://github.com/FairwindsOps/docs-engineering/blob/master/how/cert-manager.md)
```
sudermanjr:
```suggestion
```

sudermanjr (approved):
Approved once link is fixed. Thanks!!!!!!

PR #190, opened 2019-12-04
insights chart: minor fixes

rbren:


stable/insights-agent/templates/trivy/_job.yaml
```diff
@@ -6,8 +6,8 @@ volumes:
   emptyDir: {}
 containers:
 - name: trivy
-  image: quay.io/fairwinds/fw-trivy
```
makoscafee, line 9: Oops! Good catch. Not sure how we did not see this in the first place.

makoscafee (approved):
LGTM.

stable/insights-agent/templates/trivy/_job.yaml
rbren: Me too!

PR #191, opened 2019-12-04
Add stackdriver adapter chart

mjhuber:
Added a `stackdriver-metrics-adapter` chart.  This chart installs a utility to expose stackdriver metrics as external metrics in a gke cluster, making them available for HPAs to scale with.


incubator/stackdriver-metrics-adapter/values.yaml
```diff
@@ -0,0 +1,23 @@
+replicaCount: 1
```
sudermanjr, line 1: Can this run with multiple replicas?  I imagine people don't want to lose their custom metrics when an update happens.

sudermanjr (approved):
One small question inline. I think this is fine, but I haven't verified functionality personally. I know you have though, so 👍

incubator/stackdriver-metrics-adapter/values.yaml
mjhuber: yep it can work with 2 replicas

PR #193, opened 2019-12-05
insights-agent: update kube-hunter to official image

rbren:


stable/insights-agent/values.yaml
```diff
@@ -39,8 +39,8 @@ kubehunter:
   enabled: true
   schedule: "rand * * * *"
   image:
-    repository: quay.io/fairwinds/kube-hunter
-    tag: "1.0.0"
+    repository: aquasec/kube-hunter
+    tag: "501"
```
makoscafee, line 43: Probably update `kubehunter.image.repository` and `kubehunter.image.tag` in `README` as well

makoscafee (approved):
LGTM. just  one comment

stable/insights-agent/values.yaml
rbren: Good point!

PR #196, opened 2019-12-12
Update Goldilocks API Versions and Chart Testing

sudermanjr:


rbren:
Do we also need to update the vpa-installer since the current version uses `extensions/v1beta1/Deployment`?

```
unable to recognize "deploy/recommender-deployment.yaml": no matches for kind "Deployment" in version "extensions/v1beta1"
```

sudermanjr:
Unfortunately that's an issue upstream.  I'll take a look

sudermanjr:
Looks like they updated it, but didn't release. I'm looking for a commit we can pin to

sudermanjr:
I pushed something up. @rbren if you could test that I would very much appreciate it.

ejether (approved):

sudermanjr:
Just FYI, I tested on kube 1.16 in a kind cluster.

PR #199, opened 2019-12-12
 changes for astro chart 1.0.0

lucasreed:
- changed config value names starting with "controller" to start with "deployment" instead
- added permissions to coordination api for leader election
- made rbac a toplevel config value
- made replicas configurable
- added a command to the container spec to allow specification of an argument
- updated chart version to 1.0.0
- default img version to 1.5.2

I think after this is merged we can open a PR to move this chart into `stable/`

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,5 +1,6 @@
 *       @bambash @sudermanjr
 
+/incubator/astro/ @mjhuber @lucasreed
 /incubator/autospotting/ @sudermanjr @coreypobrien
 /incubator/basic-demo/ @sudermanjr
 /incubator/capsize/ @sudermanjr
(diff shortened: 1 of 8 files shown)


sudermanjr (changes requested):
Can you please put a note about the values changes when upgrading from <v1 to 1.0.0+ ? This might break any existing installs.

In addition, can you please update the CODEOWNERS file at the root of the repo to make the Astro Codeowners the same for the chart? This will be a requirement for going to stable.

mjhuber:
I incremented `image.tag` to `v1.5.2` to get an important bug fix in this update

lucasreed:
Just making it back to this. I added documentation to the README for the changed values and squashed down to a single commit with all the changes.

sudermanjr (approved):

PR #201, opened 2019-12-17
insights-agent: add option for dockerconfig to trivy

rbren:


makoscafee (approved):
LGTM... I thought should also be a way to expose `MAX_CONCURENT` variable with 5 being the default ?

rbren:
Great point!

PR #202, opened 2019-12-20
bump up kubesec timeout

rbren:
Kubesec take O(workloads) to run, and takes 1-2 minutes on a relatively small cluster. Setting to 20 minutes so we can handle larger clusters.


makoscafee (commented):
Forgot to bump chart version so linting is failing

makoscafee (approved):
LGTM. Thanks Robert

PR #203, opened 2020-01-03
fixing a bug in dnsNames edge case where a wildcard would cause a formatting issue

lucasreed:


mjhuber (approved):

ejether (approved):
TIL the `quote` filter in Go templates

PR #204, opened 2020-01-09
[helm-release-pruner] Fix service account name

rbren:


coreypobrien (changes requested):
I think what we really want here is
a) By default, the chart uses a generated name that includes the release
b) If job.serviceAccountName is specified, we use that instead and we don't generate a service account (serviceAccountName would be an override essentially to use something existing)

rbren:
I had to duplicate some of the logic in _helpers.tpl (copying the `fullname` logic over to `serviceaccount`). `template` doesn't seem to play nicely with sprig functions...a couple things I tried that didn't work:
 
In the yaml files:
```
{{ .Values.serviceAccountName | default template helm-release-pruner.fullname }}
```

In _helpers.tpl:
```
{{- define "helm-release-pruner.serviceaccount" }}
{{- if .Values.serviceAccountName }}
{{- .Values.serviceAccountName }}
{{ template "helm-release-pruner.fullname" . }}
{{- else }}
{{- end }}
```

Any ideas on how to do this better?

sudermanjr:
I usually just do it like this:

https://github.com/FairwindsOps/charts/blob/507d676c2043d2680fe9f2621bd1785103063be7/stable/goldilocks/values.yaml#L22-L24

https://github.com/FairwindsOps/charts/blob/507d676c2043d2680fe9f2621bd1785103063be7/stable/goldilocks/templates/controller-deployment.yaml#L26-L30

sudermanjr:
@rbren Not sure if you wanted to change this, or if you are good to go.

rbren:
Oops sorry, lost track of this. Will take a look at your suggestion Andy

bbensky (approved):
LGTM

rbren:
@coreypobrien looks like I'm blocked pending your :rubberstamp:

incubator/helm-release-pruner/values.yaml
```diff
@@ -9,9 +9,11 @@ job:
   backoffLimit: 3
   restartPolicy: Never
   schedule: "0 */4 * * *"
-  serviceAccountName: helm-release-pruner
   dryRun: true
   debug: false
+  serviceAccount:
+    create: true
+    # name: ExistingServiceAccountName
```
sudermanjr, line 16: Normally it's better to leave this uncommented so that you can pass a `helm lint --strict`. Just a note for future reference, non-blocking.

sudermanjr (approved):

PR #207, opened 2020-01-14
 Add pods to allowed resources in fairwinds-metrics clusterrole

dosullivan:
The chart was missing permissions for fairwinds-metrics to query pod resources, which is necessary for the terminating pods check. Without this, you'll see this error message in the pod's logs:

```
E0114 16:30:03.330531       1 terminating_pods.go:27] Error getting pods: pods is forbidden: User "system:serviceaccount:fairwinds-metrics:fairwinds-metrics-controller" cannot list resource "pods" in API group "" at the cluster scope
```


sudermanjr (approved):
Thanks!

PR #208, opened 2020-01-15
Add apiVersion to rbac-manager

fraserdarwent:
Add `apiVersion` field to `rbac-manager` Helm chart to pass `helm lint`


sudermanjr (approved):
Thanks!!

PR #211, opened 2020-01-17
Convert goldilocks icon to png

mjhuber:
Converted the goldilocks icon to png.  Helm hub does not render svg icons correctly.

image


sudermanjr (approved):
Thanks!!!!

PR #217, opened 2020-01-28
Re-run Insights-Agent jobs

baderbuddy:
This adds a helm hook to run after install/upgrade that will create a job from all of the cronjobs created instead of the helm chart creating those jobs. This way the chart can be run multiple times.


stable/insights-agent/templates/helm-hook/job.yaml
```diff
@@ -0,0 +1,22 @@
...
+spec:
+  ttlSecondsAfterFinished: 30
+  backoffLimit: {{ .Values.cronjobs.backoffLimit }}
+  template:
+    spec:
+      restartPolicy: Never
+      serviceAccountName: {{ include "insights-agent.fullname" . }}-cronjob-executor
+      containers:
+      - name: start-job-from-cronjob
+        image: bitnami/kubectl:{{ .Capabilities.KubeVersion.Major }}.{{ .Capabilities.KubeVersion.Minor }}
```
rbren, line 17: great call on versioning!

stable/insights-agent/templates/helm-hook/job.yaml
```diff
@@ -0,0 +1,22 @@
...
+        command: ["bash"]
+        args:
+          - -c
+          - |
+            kubectl get cj --selector=app=insights-agent -o wide | grep insights | cut -d' ' -f1 | xargs -I~job~ sh -c "kubectl delete job ~job~ || echo 'Does Not Exist'"
```
rbren: IIUC, this would delete the the cronjob job history, which the user might want to keep around

Maybe we could give the job a name like `$reportname-init`, and grep for that? Or add a label so we could select specifically for the init jobs before deleting them?

stable/insights-agent/templates/helm-hook/job.yaml
```diff
@@ -0,0 +1,22 @@
...
+            kubectl get cj --selector=app=insights-agent -o wide | grep insights | cut -d' ' -f1 | xargs -I~job~ kubectl create job ~job~ --from=CronJob/~job~
```
rbren: OOC, why not do `-o name` and skip the `cut`? Also why `grep insights`? Guessing the two are related.

stable/insights-agent/templates/helm-hook/job.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: batch/v1
```
rbren, line 2: Can we actually gate this job on a variable like `runJobsImmediately`? That might come in handy down the line.

stable/insights-agent/templates/helm-hook/job.yaml
baderbuddy: This will only delete the jobs that's names EXACTLY match the cronjob, so just the ones that this is creating not the ones that the CronJobs create.

stable/insights-agent/templates/helm-hook/job.yaml
baderbuddy: Whoops :)

stable/insights-agent/templates/helm-hook/job.yaml
baderbuddy: Done

stable/insights-agent/templates/helm-hook/job.yaml
baderbuddy: Because when the CronJob creates a job it appends the time at the end of the Job name.

baderbuddy:
Oh I just realized -o name includes cronjob.batch on the beginning so I'm still going to have to cut it.

baderbuddy:
`kubectl get cj --selector=app=insights-agent -o custom-columns=NAME:.metadata.name --no-headers` should work, it won't append cronjob.batch to the beginning, no weird cut needed.

stable/insights-agent/templates/helm-hook/job.yaml
rbren: 👍 thanks for the explanation!

rbren (approved):

PR #221, opened 2020-02-04
Bump versions of Insights-Agent dependencies

baderbuddy:
Increased the versions of golidlocks, fw-trivy, and workloads to the latest versions.

Run Trivy as non-privileged now


rbren:
Thanks! I think we can remove this exemption from the Polaris config:

https://github.com/FairwindsOps/charts/blob/master/stable/insights-agent/templates/polaris/configmap.yaml#L130

Are we able to set `runAsUser` instead of running as root now?

rbren (approved):

PR #182, opened 2019-11-26
Add Release watcher Chart in insight agent

makoscafee:
This PR aims at adding release watcher in insight agents. Few things to figure out before is merged

- [x]  What are the permission needed to run this job
- [x]  Question on the output format file, As the syntax used will it create output in a file?.

PR #223, opened 2020-02-05
Added RBAC-reporter chart

makoscafee:


stable/insights-agent/templates/rbac-reporter/rbac.yaml
```diff
@@ -0,0 +1,16 @@
...
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ include "insights-agent.fullname" . }}-rbac-reporter
+  labels:
+    app: insights-agent
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: edit
```
rbren: Where is this `ClusterRole` coming from? I'm noticing now we have the same `ClusterRole` in `release-watcher/rbac.yaml` - where is it defined?

We shouldn't need edit permissions for this or release-watcher, just view permissions.

stable/insights-agent/Chart.yaml
```diff
@@ -1,8 +1,9 @@
 apiVersion: v1
 description: A Helm chart to run the Fairwinds Insights agent
 name: insights-agent
-version: 0.12.1
+version: 0.12.2
```
rbren: Let's bump the minor version since we're adding a new report

stable/insights-agent/templates/rbac-reporter/rbac.yaml
makoscafee: changed for `rbac-reporter`. For the case of release watcher the discussion was [ here](https://github.com/FairwindsOps/charts/pull/182#pullrequestreview-323005723)

makoscafee: also looking at this chart rbac https://github.com/FairwindsOps/release-watcher/blob/master/chart/templates/rbac.yaml ... should we replicate from this?

stable/insights-agent/templates/rbac-reporter/_job.yaml
```diff
@@ -0,0 +1,41 @@
+{{- define "rbacreporter.job.spec" -}}
```
baderbuddy, line 1: We probably don't have to keep up defining the separate _job file since we don't have both CronJob and Job anymore.

stable/insights-agent/templates/rbac-reporter/_job.yaml
```diff
@@ -0,0 +1,41 @@
...
+restartPolicy: Never
+serviceAccountName: {{ include "insights-agent.fullname" . }}
```
baderbuddy: I for one would like to see us start using different Service Accounts when one of the jobs needs special permissions like this. Not sure if I'm being paranoid but principal of least privilege is stuck in my head :)

stable/insights-agent/templates/rbac-reporter/_job.yaml
```diff
@@ -0,0 +1,41 @@
...
+- name: output
+  emptyDir: {}
+containers:
+- name: rbac-reporter
+  image: "{{ .Values.rbacreporter.image.repository }}:{{ .Values.rbacreporter.image.tag }}"
+  imagePullPolicy: Always
+  args:
+    - cluster
+    - -u
+    - https://charts.fairwinds.com/stable
```
baderbuddy: What are these Helm repos here for?

stable/insights-agent/templates/rbac-reporter/_job.yaml
```diff
@@ -0,0 +1,41 @@
...
+  volumeMounts:
+  - name: output
+    mountPath: /output
+  resources:
+    {{- toYaml .Values.rbacreporter.resources | nindent 4 }}
+  securityContext:
+    readOnlyRootFilesystem: true
+    allowPrivilegeEscalation: false
+    runAsNonRoot: true
+    runAsUser: 1200
```
baderbuddy: Does it have to run as a specific user? If not it already can't run as root, so specifying a particular user seems unnecessary.

stable/insights-agent/values.yaml
```diff
@@ -140,3 +140,17 @@ trivy:
...
+  schedule: "rand * * * *"
+  image:
+    repository: quay.io/fairwinds/rbac-reporter
+    tag: "v1.0.0"
+  resources:
+    limits:
+      cpu: 100m
+      memory: 128Mi
+    requests:
+      cpu: 100m
```
baderbuddy: Do we need all of these resources or could we lower it down?

baderbuddy (commented):
I had a few notes below

stable/insights-agent/templates/rbac-reporter/_job.yaml
makoscafee: ops fixed

stable/insights-agent/templates/rbac-reporter/_job.yaml
makoscafee: Not sure i understand this one here. You mean `_job.yaml` and `cronjob.yaml` are not needed in this context?

stable/insights-agent/templates/rbac-reporter/_job.yaml
makoscafee: created separate service account

stable/insights-agent/templates/rbac-reporter/serviceaccount.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
```
rbren: I don't think we need a separate serviceaccount - we can just use the one that's created in serviceaccount.yaml

stable/insights-agent/templates/rbac-reporter/rbac.yaml
```diff
@@ -0,0 +1,16 @@
...
+  name: {{ include "insights-agent.fullname" . }}-rbac-reporter-view
+  labels:
+    app: insights-agent
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: view
+subjects:
+  - kind: ServiceAccount
+    name: {{ include "insights-agent.fullname" . }}-rbac-reporter
```
rbren: Once you delete the serviceaccount below, change this to `{{ include "insights-agent.fullname" . }}`

stable/insights-agent/templates/rbac-reporter/_job.yaml
rbren: Before, we used `_job.yaml` to factor out the common parts of the Job and CronJob. Now we're not creating separate Jobs (just the CronJob) so we don't really need _job.yaml anymore.

I'm good leaving this in for consistency, but we should refactor the chart to get rid of this pattern in the future.

makoscafee:
Removed `_job.yaml` and separate `serviceaccount`

rbren (approved):

PR #233, opened 2020-02-06
Bringing release-watcher to latest version in insights-agent

baderbuddy:
Bringing up version and removing the trivy exclusion for Polaris.

### DIFF
--- a/stable/insights-agent/templates/release-watcher/rbac.yaml
+++ b/stable/insights-agent/templates/release-watcher/rbac.yaml
@@ -1,14 +1,43 @@
 {{- if .Values.releasewatcher.enabled -}} 
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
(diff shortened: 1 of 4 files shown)


stable/insights-agent/templates/release-watcher/rbac.yaml
```diff
@@ -1,14 +1,43 @@
...
 metadata:
-  name: {{ include "insights-agent.fullname" . }}-release-watcher
+  name: {{ include "insights-agent.fullname" . }}-release-watcher-view
   labels:
     app: insights-agent
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: edit
+  name: view
```
rbren, line 11: Where is this `ClusterRole` coming from?

baderbuddy: View? It's a built in ClusterRole, I mimicked the changes to the release-watcher helm chart to lower the amount of permissions we're taking.

rbren: Aha! I knew I was missing something.

rbren (approved):

PR #236, opened 2020-02-11
Use orb for e2e multiple versions of kube

### LINKED ISSUE #197: [Chart Automation] Move to e2e testing that is setup by Rok8s CircleCI Orb.
The rok8s-scripts orb offers a very streamlined and maintained setup of a Kind cluster. In theory, we should be able to port the testing here over to that setup and still use the chart-test repo.

This would eliminate about 3/4 of the bash that we have hiding in this repo for setting up e2e tests, and make it easier to run e2e tests against multiple versions of Kube.

sudermanjr:
Fixes #197


sudermanjr:
Okay, so I haven't yet figured out how to make helm3 work, and the helm3 chart-testing container seems to have some bugs (and is running an old version of helm3).

So for now, this PR will implement the rok8s-orb-style e2e tests for multiple versions of kube. I will leave out the helm3 versions for now.

lucasreed (commented):
One issue I've noticed when trying to run this locally is that since the executor container was taken out, the `scripts/e2e-test.sh setup` command does not set your kubeconfig so the `kubectl` commands after kicking off the kind cluster don't work. Since we use the orb to run `scripts/e2e-test.sh test` we can't get the kubeconfig of our local kind cluster during the test step because that would screw up CI tests (that environment already has a kubeconfig set ahead of time).

One possible solution is to add this to the e2e-test script:
``` diff
diff --git a/scripts/e2e-test.sh b/scripts/e2e-test.sh
index 0cf942d..6db1886 100755
--- a/scripts/e2e-test.sh
+++ b/scripts/e2e-test.sh
@@ -20,10 +20,12 @@ set -x

OPERATION="${1:-create}"
 TEST_CLUSTER_NAME="${3:-helm-e2e}"
+KUBECONFIG=""

 setup_cluster () {
     printf "Creating cluster %s.  This could take a minute...\n" "$TEST_CLUSTER_NAME"
     kind create cluster --name="$TEST_CLUSTER_NAME" --wait=120s > /dev/null
+    KUBECONFIG=$(kind get kubeconfig-path --name "$TEST_CLUSTER_NAME")
 }

 install_hostpath_provisioner() {
```

Then, maybe in the CONTRIBUTING.md doc just mention to set your kubeconfig manually before running `scripts/e2e-test.sh test`

sudermanjr:
@lucasreed I'm a little confused as to why this is an issue. By default kind 0.6+ uses whatever your local KUBECONFIG is, or is uses `~/.kube/config`.  

When I setup the cluster locally my kubeconfig gets configured in my ~/.kube/config and then I have access to the cluster.

Are you perhaps on an old verison of Kind?

EDIT: In fact `kind get kubeconfig-path` is no longer a valid command.

lucasreed:
> @lucasreed I'm a little confused as to why this is an issue. By default kind 0.6+ uses whatever your local KUBECONFIG is, or is uses `~/.kube/config`.
> 
> When I setup the cluster locally my kubeconfig gets configured in my ~/.kube/config and then I have access to the cluster.
> 
> Are you perhaps on an old verison of Kind?
> 
> EDIT: In fact `kind get kubeconfig-path` is no longer a valid command.

Ahh, sorry for the noise. I was using 0.5.0.

sudermanjr:
I added the helm init function back into the setup for local testing.

lucasreed (approved):
LGTM, thanks @sudermanjr !

PR #238, opened 2020-02-12
Update insights-agent chart for trivy rolling scans

baderbuddy:
Updates the version of the Trivy image to scan the last N (by default 30 right now) images each run of the cronjob.


stable/insights-agent/values.yaml
```diff
@@ -136,15 +136,16 @@ releasewatcher:
 
 trivy:
   enabled: true
-  namespaceBlacklist: nil
+  namespaceBlacklist: []
   schedule: "rand */3 * * *"
   privateImages:
     dockerConfigSecret: ""
   maxConcurrentScans: 5
+  maxScansPerRun: 30
```
rbren:
```suggestion
  maxScansPerRun: 10
```

I think that'll still take ~5 minutes

stable/insights-agent/values.yaml
```diff
@@ -136,15 +136,16 @@ releasewatcher:
...
   timeout: 10800
```
rbren:
```suggestion
  timeout: 1200
```

rbren (approved):

PR #239, opened 2020-02-12
Make Trivy filesystem writeable

baderbuddy:
Trivy broke with readonly root.


rbren (approved):

sudermanjr:
Is it possible to instead mount an emptydir in the place where it needs to save files? Allowing root filesystem writes is a security issue.

rbren:
Sadly (AFAICT) trivy does not appear to be readily configurable. This reverts a recent attempt to make the FS read only (for that very reason)

sudermanjr:
Ah, bummer.

rbren:
Hmm wait, might have figured it out...

PR #240, opened 2020-02-12
make trivy read-only again

rbren:


baderbuddy (approved):
Awesome, approved with the version getting bumped over the current.

PR #241, opened 2020-02-13
Letsencrypt setup azuredns

bambash:
This PR adds support for AzureDNS in the letsencrypt-setup chart.

### DIFF
--- a/incubator/letsencrypt-setup/README.md
+++ b/incubator/letsencrypt-setup/README.md
@@ -62,6 +62,20 @@ The following table lists the configurable parameters of the chart and their def
 | `secretKey` | Key within a kubernetes secret data block that holds the secret access key |
 | `type` | Defines the type of DNS solver, should be set to `cloudflare` |
 
+### AzureDNS
+
+| Parameter | Description |
+| --------- | ----------- |
+| `clientId` | The AAD Service Principal Application ID |
+| `clientSecret` | Name of kubernetes secret that contains the secret Service Principal password |
+| `clientSecretKey` | Key within a kubernetes secret data block that holds the Service Principal password |
+| `subscriptionId` | Azure subscription where the DNS zone lives |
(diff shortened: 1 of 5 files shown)


incubator/letsencrypt-setup/README.md
```diff
@@ -62,6 +62,20 @@ The following table lists the configurable parameters of the chart and their def
...
+
+| Parameter | Description |
+| --------- | ----------- |
+| `clientId` | The AAD Service Principal Application ID |
+| `clientSecret` | Name of kubernetes secret that contains the secret Service Principal password |
+| `clientSecretKey` | Key within a kubernetes secret data block that holds the Service Principal password |
+| `subscriptionId` | Azure subscription where the DNS zone lives |
+| `tenantId` | Azure tenant where the subscription lives |
+| `resourceGroupName` | Resource group where the DNS zone lives |
+| `hostedZoneName` | DNS Zone name (ex: example.com) |
```
sudermanjr, line 75: Is it necessary to filter to a single zone per clusterissuer? I for some reason though that cert-manager could manage multiple with a single clusterissuer

sudermanjr (approved):
LGTM. I will also defer to @lucasreed since he knows this chart a lot better than me. I left one question inline, but it's a non-blocker

incubator/letsencrypt-setup/README.md
bambash: i'm not completely sure, that is a good question. I mainly followed the docs here: https://cert-manager.io/docs/configuration/acme/dns01/azuredns/

sudermanjr: Looks like you could do multiple solvers, one for each zone you wanted. either way this is fine the way it is.

lucasreed (approved):
This looks great, thanks for adding it! 👏

PR #244, opened 2020-02-27
KubeBench for Insights Agent

baderbuddy:
Added kube-bench as an option.

Bumped the app version number, I bumped the chart version only as a patch change because we're already doing a new minor release for the next version of Insights anyways.

Some of the RBAC we didn't have gated, so I added a few IFs.

The CronJob I wanted to make use of the _cronjob template that already exists but didn't want to continue the pattern of the _job file because there's not 2 separate job/cronjobs.

Depending on `kubebench.deploymentMode` this can deploy in one of three different ways. Kube-Bench has to run with hostPID and as root.

### DIFF
--- a/stable/insights-agent/Chart.yaml
+++ b/stable/insights-agent/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 description: A Helm chart to run the Fairwinds Insights agent
 name: insights-agent
 version: 0.14.0
-appVersion: 0.7.1
+appVersion: 0.8.0
 maintainers:
   - name: rbren
   - name: makoscafee
--- a/stable/insights-agent/Chart.yaml
+++ b/stable/insights-agent/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to run the Fairwinds Insights agent
 name: insights-agent
-version: 0.13.1
+version: 0.14.0
 appVersion: 0.7.1
 maintainers:
   - name: rbren
--- a/stable/insights-agent/Chart.yaml
+++ b/stable/insights-agent/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
 description: A Helm chart to run the Fairwinds Insights agent
 name: insights-agent
-version: 0.13.0
-appVersion: 0.6.0
+version: 0.13.1
+appVersion: 0.7.1
 maintainers:
   - name: rbren
   - name: makoscafee
--- a/stable/insights-agent/README.md
+++ b/stable/insights-agent/README.md
@@ -54,6 +54,9 @@ Parameter | Description | Default
 `{report}.resources` | CPU/memory requests and limits for the report |
 `{report}.image.repository` | Repository to use for the report image |
 `{report}.image.tag` | Image tag to use for the report |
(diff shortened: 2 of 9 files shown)


stable/insights-agent/values.yaml
```diff
@@ -148,6 +148,34 @@ rbacreporter:
       cpu: 10m
       memory: 64Mi
 
+kubebench:
+  enabled: true
+  schedule: "rand * * * *"
+  deploymentMode: "daemonsetMaster"
```
rbren: per thread w/ EJ, let's move to `cronjob` as the default

stable/insights-agent/templates/trivy/rbac.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.trivy.enabled -}}
```
rbren, line 1: good catch!

stable/insights-agent/templates/kube-bench/cronjob.yaml
```diff
@@ -0,0 +1,100 @@
...
+- name: etc-kubernetes
+  hostPath:
+    path: "/etc/kubernetes"
+    type: Directory
+- name: usr-bin
+  hostPath:
+    path: "/usr/bin"
+    type: Directory
+containers:
+- image: "{{ .Values.kubebench.image.repository }}:{{ .Values.kubebench.image.tag }}"
```
rbren: Can we factor out this container? Looks like it's mostly (entirely?) similar between cronjob and daemonset

stable/insights-agent/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 description: A Helm chart to run the Fairwinds Insights agent
 name: insights-agent
-version: 0.13.0
-appVersion: 0.6.0
+version: 0.13.1
```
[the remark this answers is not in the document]
baderbuddy: My best guess :) I figured since 0.13 never got released with a version of Insights it seemed safe to re-use that minor version. I was going back and forth on it though so I'd easily bump the minor version.

stable/insights-agent/README.md
```diff
@@ -54,6 +54,9 @@ Parameter | Description | Default
 `{report}.resources` | CPU/memory requests and limits for the report |
 `{report}.image.repository` | Repository to use for the report image |
 `{report}.image.tag` | Image tag to use for the report |
+`kubebench.deploymentMode` | Changes the way this plugin is deployed, either `cronjob` where it will run a single pod on the `schedule` that will pull the data necessary from a single node and report that back to Insights. `daemonset` which deploys a daemonset to the cluster which gathers data then a cronjob will gather data from each of those pods. `daemonsetMaster`  is the same as `daemonset` except the daemonset can also run on masters. | `daemonsetMaster`
```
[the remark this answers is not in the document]
baderbuddy: Good call

baderbuddy:
Made some changes after feedback from rbren

stable/insights-agent/Chart.yaml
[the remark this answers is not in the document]
rbren: Ah yeah just saw your comment above. I'd say minor just to keep it semver-y, but no big deal since Insights still recommends `0.12.*`

stable/insights-agent/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 description: A Helm chart to run the Fairwinds Insights agent
 name: insights-agent
-version: 0.13.0
-appVersion: 0.6.0
+version: 0.14.0
+appVersion: 0.7.1
```
rbren: We haven't been 100% consistent on this, but I think this will go out with release 0.8.0

rbren (approved):

PR #246, opened 2020-03-02
[fairwinds-metrics] The secret should not be created if the listener is not enabled

sudermanjr:


bbensky (approved):
LGTM

mjhuber (approved):
👍 thanks for the fix!

PR #247, opened 2020-03-02
Increase kube-hunter log level for insights-agent

baderbuddy:


rbren (approved):

stable/insights-agent/templates/kube-hunter/_job.yaml
```diff
@@ -15,7 +15,7 @@ containers:
     - |
       python ./kube-hunter.py \
         --pod \
-        --log none \
+        --log WARNING \
```
rbren, line 18: Do we want to either (a) bump this up to `INFO` or (b) parameterize it?

PR #260, opened 2020-03-12
Move workload to its own service account

baderbuddy:
I noticed workloads were missing a permission if not installed with one of the other plugins so I'm going ahead and moving it to its own service account to make it easier to catch these overlaps.


rbren:
Thanks Bader! We've got an open bug to address this for all plugins.

rbren (approved):

makoscafee (approved):

PR #265, opened 2020-03-17
Add separate service account to each agent

makoscafee:


stable/insights-agent/templates/kubesec/serviceaccount.yaml
```diff
@@ -0,0 +1,8 @@
+{{- if .Values.kubesec.enabled -}}
```
baderbuddy: I'd say let's not bother with an individual service account when there's no RBAC, but I'd be happy to hear other thoughts on this.

stable/insights-agent/templates/kube-bench/serviceaccount.yaml
```diff
@@ -0,0 +1,8 @@
+{{- if .Values.kubebench.enabled -}}
```
baderbuddy: Whatever is decided with kubesec below would apply here too.

baderbuddy (commented):
Looks pretty good, have you tested it out? Also I have a question about whether or not it's okay to have a generic service account when there's no specific permissions granted.

stable/insights-agent/templates/kubesec/serviceaccount.yaml
rbren: I think we definitely need some RBAC for kubesec, it `kubectl get`s a bunch of controllers.

stable/insights-agent/templates/kube-bench/serviceaccount.yaml
rbren: I agree - we can probably drop the SA here (as well as `serviceAccountName` in the podspec)

stable/insights-agent/templates/kubesec/serviceaccount.yaml
makoscafee: `kubesec` uses/gets/list these resources `deployments,statefulsets,daemonsets,namespaces`

rbren:
@baderbuddy IIUC if we don't specify a particular service account, it'll use the "default service account", which is set cluster-to-cluster. I think we should probably specify in that case

stable/insights-agent/templates/kube-hunter/rbac.yaml
```diff
@@ -0,0 +1,49 @@
+{{- if .Values.kubehunter.enabled -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "insights-agent.fullname" . }}-kube-hunter
+  labels:
+    app: insights-agent
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
```
rbren: I think we can delete this ClusterRole and ClusterRoleBinding - kube-hunter doesn't seem to need any RBAC. Let's just use an empty service account

stable/insights-agent/templates/kube-bench/cronjob.yaml
```diff
@@ -2,7 +2,6 @@
 {{- $_ := set . "Label" "kube-bench" -}}
 {{- $_ := set . "Config" .Values.kubebench -}}
 restartPolicy: Never
-serviceAccountName: {{ include "insights-agent.fullname" . }}
```
rbren, line 5: Let's create and use an empty service account here, like in kube-hunter

rbren (approved):

PR #266, opened 2020-03-23
Correct if on insights agent chart kubebench

baderbuddy:


rbren (approved):
Good catch! was kubehunter not enabled?

PR #268, opened 2020-03-25
Fix goldilocks rbac

rbren:
copied from here: https://github.com/FairwindsOps/goldilocks/blob/master/hack/manifests/controller/clusterrole.yaml


baderbuddy (approved):
LGTM

PR #271, opened 2020-03-30
removing runAsUser from deployment and increase rbac-manager version

### LINKED ISSUE #270: [rbac-manager] Hard-coded runAsUser parameter is incompatible with Openshift 4
**Describe the bug**
I cannot install rbac-manager using the helm chart.  The deployment tries to start pods and fails:

> Error creating: pods "rbac-manager-78cbdd9fdb-" is forbidden: unable to validate against any security context constraint: [spec.containers[0].securityContext.securityContext.runAsUser: Invalid value: 1200: must be in the ranges: [1000570000, 1000579999]]

**To Reproduce**
helm repo add fairwinds-stable https://charts.fairwinds.com/stable
helm install fairwinds-stable/rbac-manager --name rbac-manager --namespace rbac-manager
kubectl describe rs -n rbac-manager

**Expected behavior**
Pods would start up and run as any user in the acceptable range, or I would be instructed to create a service account and add it to a securityContextConstraint using `oc adm policy add-scc-to-user X`.

**Environment (please complete the following information):**
 - Helm Version: v3.1.1
 - Kubernetes Version v1.16.2 (Openshift 4.3.8)

lucasreed:
fix #270

Corresponds with this change in the rbac-manager repo: https://github.com/FairwindsOps/rbac-manager/pull/127

### DIFF
--- a/stable/rbac-manager/Chart.yaml
+++ b/stable/rbac-manager/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: rbac-manager
-version: 1.5.2
-appVersion: 0.9.2
+version: 1.5.3
+appVersion: 0.9.3
 description: A Kubernetes operator that simplifies the management of Role Bindings and Service Accounts.
 icon: https://raw.githubusercontent.com/FairwindsOps/charts/master/stable/rbac-manager/icon.png
 keywords:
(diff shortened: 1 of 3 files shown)


sudermanjr (approved):
Thanks!

PR #272, opened 2020-04-01
[helm-release-pruner] Move to stable and update for helm3

sudermanjr:
Also started using helm-docs for this chart.


robertlucido (approved):
🐎

PR #273, opened 2020-04-03
update release-watcher image

makoscafee:


stable/insights-agent/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to run the Fairwinds Insights agent
 name: insights-agent
-version: 0.16.4
+version: 0.16.5
```
rbren: Let's bump minor - no need to pull this into the live version of insights.

rbren (approved):

PR #275, opened 2020-04-10
updating notes for dashboard component

shraykay:
get pods with the original label selector providers access to the controller, instead of the dashboard


sudermanjr (approved):

sudermanjr:
Thanks!

PR #279, opened 2020-04-17
bump image tag for release-watcher and update appversion

makoscafee:


stable/insights-agent/values.yaml
```diff
@@ -129,7 +129,7 @@ releasewatcher:
   timeout: 300
   image:
     repository: quay.io/fairwinds/release-watcher
-    tag: "v1.7.3"
+    tag: "v1.7.4"
```
baderbuddy, line 132: Most of our images we lock at the minor version, is there a reason we're specifying the patch version here?

baderbuddy (approved):
Looks good to me, just had a question about the image version

stable/insights-agent/values.yaml
makoscafee: Not really sure... i have seen this rule apply mostly to `insight-plugins` for external [images](https://github.com/FairwindsOps/charts/blob/a71893d1eabb5569560c7e3daed0b9708905177e/stable/insights-agent/values.yaml#L89) looks like we lock to patch version. But we can standardize this.

PR #281, opened 2020-04-28
[Goldilocks] Fixing logic to disable RBAC for disabled components

sudermanjr:


rbren (approved):
curious why the `eq X true` instead of just `X`

sudermanjr:
> curious why the `eq X true` instead of just `X`

That's an excellent question. I seem to remember having trouble with that, but I went back and tried it again and this is much cleaner.

rbren (approved):

PR #285, opened 2020-04-30
update chart to use non-deprecated api version for CRD

### LINKED ISSUE #284: [rbac-manager] update chart to use non-deprecated api version for CRD
The CRD definition `apiextensions.k8s.io/v1beta1` is deprecated in Kubernetes 1.16, yet it is still in use by the rbac-manager helm chart. 

Change `apiextensions.k8s.io/v1beta1` to  `apiextensions.k8s.io/v1` on the first line of the CRD.
https://github.com/FairwindsOps/charts/blob/master/stable/rbac-manager/templates/customresourcedefinition.yaml#L1

crmejia:
closes #284

### DIFF
--- a/stable/rbac-manager/Chart.yaml
+++ b/stable/rbac-manager/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: rbac-manager
-version: 1.5.3
+version: 1.5.4
 appVersion: 0.9.3
 description: A Kubernetes operator that simplifies the management of Role Bindings and Service Accounts.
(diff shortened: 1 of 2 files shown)


sudermanjr:
The e2e tests are failing on the upgrade path. My guess is that the straight upgrade will not work, and we will need to write an upgrade procedure and bump a major(or minor at least) version of this chart.

sudermanjr:
Since this will never pass on 1.15, which is where a lot of clusters are still at, I'm going to close this. We will open issues on both rbac-manager and this repo to investigate the upgrade and change path in the coming quarters.

PR #300, opened 2020-05-18
Update to Polaris 1.0

rbren:
**Why This PR?**
Polaris 1.0 is shipping with some breaking changes.

**Changes**
In addition to changes needed for Polaris 1.0, there are some chart changes:
* RBAC has been simplified to remove duplication
* `config` now uses the built-in Polaris config by default
* `ingress` is now attached to the dashboard values
* only a single `image` is specified for the entire chart

**Checklist:**


rbren:
I should note: I've tested both the dashboard and webhook, with and without a custom config specified.

baderbuddy (approved):
This looks great to me! :)

PR #309, opened 2020-06-11
Expose dns policy

makoscafee:
**Why This PR?**
This PR adds a way for a user to specify DNS policy that should be used in a job.

Fixes #

**Changes**

* Added DNS policy field in cronjobs values
* Added the documentation for it.

**Checklist:**

### DIFF
--- a/stable/insights-agent/README.md
+++ b/stable/insights-agent/README.md
@@ -52,6 +52,7 @@ Parameter | Description | Default
 `cronjobs.successfulJobHistoryLimit` | Number of successful jobs to keep in history for each report | 2
 `cronjobs.nodeSelector` | Node selector to use for cronjobs | null
 `cronjobs.runJobsImmediately` | Run each of the reports immediately upon install of the Insights Agent | true
+`cronjobs.dnsPolicy` | Add pod DNS policy. If dnsPolicy is not explicitly specified, then “ClusterFirst” is used |
 `{report}.enabled` | Enable the report type |
 `{report}.schedule` | Cron expression for running the report | `rand * * * *`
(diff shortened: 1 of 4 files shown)


stable/insights-agent/values.yaml
```diff
@@ -23,6 +23,7 @@ cronjobs:
   successfulJobsHistoryLimit: 2
   nodeSelector: null
   runJobsImmediately: true
+  dnsPolicy:
```
rbren: I think maybe we should comment this out? Not sure how helm would interpret this as-is

stable/insights-agent/README.md
```diff
@@ -52,6 +52,7 @@ Parameter | Description | Default
 `cronjobs.successfulJobHistoryLimit` | Number of successful jobs to keep in history for each report | 2
 `cronjobs.nodeSelector` | Node selector to use for cronjobs | null
 `cronjobs.runJobsImmediately` | Run each of the reports immediately upon install of the Insights Agent | true
+`cronjobs.dnsPolicy` | Add pod DNS policy. If dnsPolicy is not explicitly specified, then “ClusterFirst” is used |
```
rbren: Let's get rid of `If dnsPolicy is not explicitly specified, then “ClusterFirst” is used` - I think this might be configured on a per-cluster basis

makoscafee: sure let me remove it ... but I think `Default` value is the one that follows the `node` [configuration.](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy)

rbren (approved):

PR #310, opened 2020-06-15
Ssk/make readonly

shaswa:
**Why This PR?**
This PR makes kube-hunter write to the mounted FS instead of the root FS 

**Changes**

* Added `securityContext ` to kube-hunter 

**Checklist:**

* [x]  Any new values are backwards compatible and/or have sensible default.


stable/insights-agent/templates/kube-hunter/_job.yaml
```diff
@@ -25,6 +27,12 @@ containers:
   volumeMounts:
   - name: output
     mountPath: /output
+  - name: tmp
+    mountPath: /tmp
```
makoscafee, line 31: why did we mount`/temp` here?

makoscafee (approved):
LGTM. just had one question

stable/insights-agent/templates/kube-hunter/_job.yaml
baderbuddy: In Trivy we had to do the same thing. kube-hunter must write files to /tmp, and when we make the root filesystem read only we can't write to tmp anymore if it's part of the root filesystem. But by creating an emptyDir volume (which is writeable) and mounting it to /tmp, then /tmp is writeable again.

baderbuddy (approved):
This looks good to me.

shaswa:
Thanks you, Barnie and Bader!

stable/insights-agent/templates/kube-hunter/_job.yaml
makoscafee: makes sense

PR #314, opened 2020-06-29
Sc/add pluto

transient1:
**Why This PR?**
Adding Pluto to the family of tools installed with Insights.

Fixes #

**Changes**

* Add Pluto cronjob

**Checklist:**

### DIFF
--- a/stable/insights-agent/Chart.yaml
+++ b/stable/insights-agent/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to run the Fairwinds Insights agent
 name: insights-agent
-version: 0.19.2
+version: 0.20.0
 appVersion: 1.4.0
 maintainers:
   - name: rbren
--- a/stable/insights-agent/Chart.yaml
+++ b/stable/insights-agent/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
 description: A Helm chart to run the Fairwinds Insights agent
 name: insights-agent
-version: 0.20.0
-appVersion: 1.4.0
+version: 0.21.0
+appVersion: 1.6.2
 maintainers:
   - name: rbren
   - name: makoscafee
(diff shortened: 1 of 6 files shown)


stable/insights-agent/templates/pluto/rbac.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - 'secrets'
+      - 'configmaps'
+    verbs:
+      - 'get'
+      - 'list'
+      - 'watch'
+  - apiGroups:
+      - 'autoscaling.k8s.io'
+    resources:
+      - 'verticalpodautoscalers'
```
rbren: Looks like this RBAC was copied from Goldilocks - I think a better RBAC profile is probably the one release-watcher has: https://github.com/FairwindsOps/charts/blob/master/stable/insights-agent/templates/release-watcher/rbac.yaml

That gives
* cluster-view
* read access to secrets

stable/insights-agent/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to run the Fairwinds Insights agent
 name: insights-agent
-version: 0.19.1
+version: 0.19.2
```
rbren: Can we make this `0.20.0`? Usually if we're adding new features we bump the minor version.

rbren (approved):
🎉  Thanks Stevie! A couple minor comments but looks great otherwise.

Can't wait to get this in!

stable/insights-agent/templates/pluto/rbac.yaml
transient1: Got it. Thanks for catching that. I also see that release-watcher is using a newer apiVersion too so I updated those as well. I adopted the use of the `view` clusterrole for the pluto serviceaccount, and created a separate clusterrole for being able to view secrets. I kept namespaces (because Helm 3 is namespace aware so I thought it might be needed) and configmaps as the Pluto team said that was necessary.

stable/insights-agent/Chart.yaml
transient1: Done.

rbren (approved):
Thanks Stevie!

rbren:
(made a couple fixes to version numbers after Barnie's change)

rbren:
Going to merge this

PR #321, opened 2020-07-16
Update version of polaris

baderbuddy:
**Why This PR?**
Update the version of polaris to 1.1 for Insights

Fixes #

**Changes**

* Update Polaris to version 1.1

**Checklist:**

### DIFF
--- a/stable/insights-agent/templates/polaris/configmap.yaml
+++ b/stable/insights-agent/templates/polaris/configmap.yaml
@@ -14,7 +14,7 @@ data:
       memoryRequestsMissing: warning
       memoryLimitsMissing: warning
       tagNotSpecified: danger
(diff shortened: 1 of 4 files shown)


stable/insights-agent/templates/polaris/configmap.yaml
```diff
@@ -8,56 +8,25 @@ metadata:
     app: insights-agent
 data:
   config.yaml:  |
```
rbren, line 10: I'd want to double check, but I think we can remove `config.yaml` now so that we just use the version that comes with Polaris

We should also cross-check the exemptions we have here with what's in Polaris...they fell out of sync at some point

rbren (approved):

PR #331, opened 2020-07-24
[insights-agent] refactor cronjobs

rbren:
**Why This PR?**
insights-agent creates two objects per report: a CronJob, as well as Job with the same details, so that the CJ runs immediately. To support this, we factored out the common parts into a `_job.yaml`, which was included in both `cronjob.yaml` and `job.yaml`

A while ago we changed from having Job specs defined directly in the chart, to using `k create job --from=cronjob`. So this factoring out is no longer necessary.

This PR moves back to a more straightforward way of doing things.

I'm not convinced it's strictly better. It's a little easier to read, since the `cronjob.yaml` actually _looks_ like a CronJob. But there's a bit more boilerplate involved, since it needs to import cronjob-spec, job-spec, and job-template-spec.

Open to any thoughts/ideas here.

**Checklist:**


baderbuddy:
Oh that is fantastic!

rbren:
Managed to factor out a few more things:
* resources
* image + pull policy
* container name (applied to all reports, this would make container-name consistently equal to report-name)
* security context (will need to be overridden in certain circumstances, but it'll be nice to know who's using non-standard security context)

rbren:
This should be ready to go!

The downstream changes to our install yaml are here: https://github.com/FairwindsOps/Insights/compare/rb/refactor-agent

Most of it is rearranging fields and other trivial things (like changing container names). There are a few substantive changes though:
* some containers are now getting `runAsUser: 10324`
* some are now getting `privileged: false`

rbren:
Also worth noting, there's still this big chunk of boilerplate at the top:
```yaml
apiVersion: batch/v1beta1
kind: CronJob
metadata:
  {{- include "metadata" .Label }}
spec:
  {{ include "cronjob-spec" . | nindent 2 | trim }}
  jobTemplate:
    spec:
      {{ include "job-spec" . | nindent 6 | trim }}
      template:
        spec:
          {{ include "job-template-spec" . | indent 10 | trim }}
```

we could factor that out into its own thing, but I kind of like that each `cronjob.yaml` still _looks_ like a CronJob. And any copy/paste issues will probably lead to a failed build rather than a bug. Open to suggestions though.

baderbuddy (approved):
This looks so good!

PR #337, opened 2020-08-07
Add opa to Insights Agent

baderbuddy:
**Why This PR?**

Fixes #

**Changes**

* Adds a plugin for opa
*

**Checklist:**


stable/insights-agent/templates/opa/annotation-blocked.yaml
```diff
@@ -0,0 +1,22 @@
+{{- if .Values.opa.enabled -}}
+apiVersion: insights.fairwinds.com/v1
```
rbren: I like that you built a handful of test checks!

We should probably provide these to folks in another context, rather than automatically including them with an OPA install. We're anticipating OPA users are mostly going to want to do their own thing, but it'd be good to provide a handful of examples and pre-built checks. Let's keep them in our pocket for now.

stable/insights-agent/templates/opa/rbac.yaml
```diff
@@ -0,0 +1,40 @@
...
+    app: insights-agent
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: view
+subjects:
+  - kind: ServiceAccount
+    name: {{ include "insights-agent.fullname" . }}-opa
+    namespace: {{ .Release.Namespace }}
+{{- end -}}
```
rbren, line 73: we should add the ability to add additional cluster roles for this ServiceAccount, in case they want to check a resource that hasn't been aggregated to cluster-view

stable/insights-agent/templates/opa/label-required.yaml
```diff
@@ -0,0 +1,21 @@
...
+  name: must-have-label
+spec:
+  output:
+    title: "Label is missing"
+    severity: 0.1
+    remediation: "Add the label"
+    category: "Reliability"
+  rego: |
+         package fairwinds
+         labelrequired[description] {
```
rbren: how does one return, e.g. a `description` and a `severity`? Is that possible?

baderbuddy: Yeah, I suppose none of the custom checks I provided show that. We'll have to figure out if that's appropriate for just documentation or if we can contrive an example where it would be good to actually demonstrate it.

So regardless of what that variable is named. If you return an Object instead of a string, then that object can have severity, description, category, and title. I believe all of them will override whatever is in the check.

stable/insights-agent/templates/opa/rbac.yaml
baderbuddy: Yeah that's probably a good idea. I always struggle with that. The right way to do it seems to be to aggregate into view (like we're doing here), but not everyone seems to follow that. Also they could legitimately want to run an opa check on secrets for some reason.

stable/insights-agent/crds/instance.yaml
```diff
@@ -0,0 +1,62 @@
+apiVersion: apiextensions.k8s.io/v1
```
makoscafee: `values.opa.enabled`?

stable/insights-agent/crds/check.yaml
```diff
@@ -0,0 +1,59 @@
+apiVersion: apiextensions.k8s.io/v1
```
makoscafee: `values.opa.enabled`?

stable/insights-agent/templates/opa/rbac.yaml
```diff
@@ -0,0 +1,74 @@
...
+    app: insights-agent
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ include "insights-agent.fullname" . }}-checks-edit
+subjects:
+  - kind: ServiceAccount
+    name: {{ include "insights-agent.fullname" . }}-opa
+    namespace: {{ .Release.Namespace }}
+{{- end -}}
```
rbren: I'm not seeing the start for this `{{end}}`

stable/insights-agent/templates/opa/rbac.yaml
```diff
@@ -0,0 +1,74 @@
...
+  - 'customchecks'
+  - 'customcheckinstances'
+  verbs:
+  - 'get'
+  - 'list'
+---
+apiVersion: rbac.authorizations.k8s.io/v1beta1
+kind: Role
+metadata:
+  name: {{ ionclude "insights-agent.fullname" . }}-checks-edit
```
rbren: `ionclude`

rbren (approved):

PR #353, opened 2020-08-26
Polaris Dependency updates

baderbuddy:
**Why This PR?**
Needed for the new version of Polaris with updated dependencies.

Fixes #

**Changes**

* New job to refresh certificates.
*

**Checklist:**


stable/polaris/templates/webhook.configuration.yaml
```diff
@@ -0,0 +1,64 @@
...
+      port: 443
+  failurePolicy: Ignore
+  matchPolicy: Exact
+  name: polaris.fairwinds.com
+  namespaceSelector:
+    matchExpressions:
+    - key: control-plane
+      operator: DoesNotExist
+  objectSelector: {}
+  rules:
```
rbren, line 29: can we make `rules` configurable, with this as the default? That way folks can decide what they want to monitor, and can add any extra controllers necessary

rbren (commented):
Looks great!

I think we just want to hold off on merging this until the new version of Polaris is ready

I'm thinking this should be a major version change for both Polaris and the Chart, but could be convinced otherwise

sudermanjr:
You'll need to move the chart version to get the linter to pass

stable/polaris/README.md.gotmpl
```diff
@@ -0,0 +1,29 @@
...
+
+### Dashboard
+```
+helm repo add fairwinds-stable https://charts.fairwinds.com/stable
+helm install polaris fairwinds-stable/polaris --namespace polaris
+```
+
+### Webhook
+
+A valid TLS certificate is required for the Polaris Validating Webhook. If you have cert-manager installed in your cluster then the helm install below will work. If you don't use cert-manager and don't want to install it then you can supply a CA Bundle with the `webhook.caBundle` parameter and create a TLS secret and pass the name of that secret as `webhook.secretName`
```
rbren: can we bullet this out a bit? i.e.
```md
If you don't use cert-manager, you'll need to:
* Supply a CA Bundle using the `webhook.caBundle` parameter
* create a TLS secret in your cluster
* pass in the name of that TLS secret using the `webhook.secretName` parameter
```

Also, is there a particular version of cert-manager we need?

rbren (approved):

stable/polaris/README.md.gotmpl
baderbuddy: 0.11 through 1.0 should all work just fine. At some point we should bump the API version of the objects we're using.

PR #356, opened 2020-09-02
Add support for selectors on http01 solver.

robertlucido:
**Why This PR?**
The `http01` solver supports the same selector fields as `dns01` and we have a case where it's necessary to use them in combination with a particular version of `cert-manager`.

**Changes**

* Support `selector` fields in the `http01` solver
* (chart structure) Use common `selector` template/helper for each `solver`
* (chart structure) Use common `solver` template for each clusterissuer
  * An `azuredns` type had been added to primary issuer but not `selfsigned`

**Checklist:**


sudermanjr:
You might need to bump this - https://github.com/FairwindsOps/charts/blob/94854dff86f8b28623ea4c2dd13aa4d478905ac0/incubator/letsencrypt-setup/Chart.yaml#L4

It controls the CRDs that get installed in the e2e tests. I'm guessing the fields you are setting don't exist in 0.11

robertlucido:
> You might need to bump this -
> 
> https://github.com/FairwindsOps/charts/blob/94854dff86f8b28623ea4c2dd13aa4d478905ac0/incubator/letsencrypt-setup/Chart.yaml#L4
> 
> It controls the CRDs that get installed in the e2e tests. I'm guessing the fields you are setting don't exist in 0.11

It was just a template error

sudermanjr (approved):

PR #360, opened 2020-09-19
[vpa] Removed empty 'required' field from CRD

chgl:
**Why This PR?**

When installing the vpa using this chart, the `required: []` field from the `verticalpodautoscalers` is removed by Kubernetes before persisting the CRD. This causes issues when using GitOps tools such as ArgoCD because the CRD manifest in git still contains the `required: []` but the live manifest does not. Thereby making the chart appear out-of-sync.

**Changes**

* delete the `required: []` line

**Checklist:**

### DIFF
--- a/stable/vpa/Chart.yaml
+++ b/stable/vpa/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: vpa
 description: A Helm chart for Kubernetes Vertical Pod Autoscaler
 type: application
-version: 0.1.0
+version: 0.1.1
 appVersion: 0.8.0
 maintainers:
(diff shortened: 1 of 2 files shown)


sudermanjr (approved):
Seems fine to me.  Thanks!

I will admit, this seems like a very serious failing of the CD tool and its ability to compare yaml.

chgl:
Thanks! I agree that an empty array value and a lack of key are semantically the same. Something like an empty object `{}` as a value vs. a lack of key are not the same though. From a `git diff` perspective the former is different as well. Then again, I don't know the internals of argocd so it's hard to say if it's really a failing or a tradeoff.

PR #362, opened 2020-09-25
Fix imagePullSecrets

### LINKED ISSUE #346: rbac-manager fails when image pulled from private repo
**Describe the bug**
getting error :getting Error: UPGRADE FAILED: error validating “”: error validating data: ValidationError(Deployment.spec.template.spec.containers[0]): unknown field “imagePullSecrets” in io.k8s.api.core.v1.Container" when install from private (seen it in 1.5.4,1.5.5,1.5.6)

**To Reproduce**
any install of rbac-manager using private image repo

possible fix is the indentation of imagePullSecrets has to be in line with the container spec as shown here(https://kubernetes.io/docs/concepts/containers/images/).
https://github.com/FairwindsOps/charts/blob/8bdc696dcc8634b8fe3fd69f867c2d4d1f5858ab/stable/rbac-manager/templates/deployment.yaml#L38

wiegandf:
**Why This PR?**
This enables rbac-manager helm chart to work with private docker registries.

Fixes #346 

**Changes**

* Bugfix: Move imagePullSecrets to Pod spec. See https://kubernetes.io/docs/concepts/containers/images/#referring-to-an-imagepullsecrets-on-a-pod 

**Checklist:**

### DIFF
--- a/stable/rbac-manager/values.yaml
+++ b/stable/rbac-manager/values.yaml
@@ -3,6 +3,7 @@ image:
   tag: v0.9.4
   pullPolicy: Always
   imagePullSecrets: []
+#   - myRegistryKeySecretName
 
 resources:
   requests:
--- a/stable/rbac-manager/Chart.yaml
+++ b/stable/rbac-manager/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: rbac-manager
-version: 1.5.6
+version: 1.5.7
 appVersion: 0.9.4
 description: A Kubernetes operator that simplifies the management of Role Bindings and Service Accounts.
 icon: https://raw.githubusercontent.com/FairwindsOps/charts/master/stable/rbac-manager/icon.png
--- a/stable/rbac-manager/templates/deployment.yaml
+++ b/stable/rbac-manager/templates/deployment.yaml
@@ -27,6 +27,12 @@ spec:
     {{- end }}
     spec:
       serviceAccountName: {{ template "rbac-manager.fullname" . }}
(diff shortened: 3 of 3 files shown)


stable/rbac-manager/values.yaml
```diff
@@ -3,6 +3,7 @@ image:
   tag: v0.9.4
   pullPolicy: Always
   imagePullSecrets: []
+#   - myRegistryKeySecretName
```
sudermanjr, line 6: This seems like an unnecessary comment here.

sudermanjr (changes requested):
Thanks for adding this!

Could you please add this value to the table in the README for the chart?

stable/rbac-manager/values.yaml
wiegandf: It helps so that you don't have to read the manifest. Some helm charts include the whole yaml which means you'd have to add `- name: myRegistryKeySecretName`

I don't have a strong opinion on this, I can remove it if you want

wiegandf:
It was already added in https://github.com/FairwindsOps/charts/pull/294/files. I would have named it pullSecrets instead of imagePullSecrets but wanted to keep the variable as it is.

sudermanjr:
> It was already added in https://github.com/FairwindsOps/charts/pull/294/files. I would have named it pullSecrets instead of imagePullSecrets but wanted to keep the variable as it is.

that's my bad for not reading closely enough, thanks for correcting me!

sudermanjr (approved):
Looks great. Thanks!!

PR #365, opened 2020-09-25
Add Insights CI checks

rbren:
**Why This PR?**
This runs the Insights CI checks against the insights-agent and insights-admission charts

Also removes a vulnerable python image discovered via Insights CI.

**Checklist:**


fairwinds-insights.yaml
```diff
@@ -0,0 +1,11 @@
+options:
```
sudermanjr, line 3: Is there any way we could keep this file somewhere else, like in `.circleci` or as a `.` file?

stable/insights-admission/values.yaml
```diff
@@ -144,4 +144,4 @@ test:
   enabled: false
   image:
     repository: "python"
-    tag: "3.6"
+    tag: "3.8-alpine"
```
sudermanjr, line 160: You'll need to run helm-docs for this change

.circleci/config.yml
```diff
@@ -113,6 +130,7 @@ workflows:
           requires:
             - lint-scripts
             - lint-charts
+            - insights
```
sudermanjr: If insights fails, you want to not run end-to-end tests?  That seems like a lot of work for you! :-D

sudermanjr: also, to be certain, this won't disrupt other builds, right. Since you have the insights config in there, it will just pass if they aren't configured?

.circleci/config.yml
```diff
@@ -74,6 +76,19 @@ jobs:
...
             git fetch ro master
             ct lint --config scripts/ct.yaml --chart-yaml-schema scripts/schema.yaml --debug
+
+  insights:
+    docker:
+      - image: quay.io/reactiveops/ci-images:v11-stretch
+    steps:
+      - checkout
+      - setup_remote_docker:
+          version: 18.09.3
```
sudermanjr: Curious why you're pinning the version here. That's not something I've seen in any of our other stuff.

baderbuddy: We probably don't need it in this case. At one point I had done that in order to use a newer version of Docker (Circle doesn't default to the latest)

fairwinds-insights.yaml
rbren: 🤔 not currently, but added a `TODO` and created an issue in Insights to make this configurable

.circleci/config.yml
```diff
@@ -74,6 +76,19 @@ jobs:
...
+      - run:
+          name: Run Insights Checks
+          command: |
+            curl https://insights.fairwinds.com/v0/insights-ci.sh | bash
```
baderbuddy: We should probably go ahead and do the -L

.circleci/config.yml
rbren: Good point. Insights should always exit 0 (we'll have a separate status check in GH to pass/fail), but might as well not wait for it.

.circleci/config.yml
rbren: Good catch - copypasta

baderbuddy (approved):
LGTM

.circleci/config.yml
rbren: 🤦

sudermanjr (changes requested):
You'll have to update from master and update the version again to fix the helm-docs annoyingness

.circleci/config.yml
rbren: Apparently you have to pin: https://circleci.com/docs/2.0/building-docker-images/#docker-version

sudermanjr: I don't see there where it says you have to pin. I see where it says you "can" pin. 

In the orb and every other CI file I have it is just `- setup_remote_docker`

rbren: Ahhh I misunderstood, I tried to just remove the patch version. Done.

sudermanjr (approved):

PR #366, opened 2020-09-28
Fix webhook

dza89:
Signed-off-by: Daan Potter <dhapotter@gmail.com>

**Why This PR?**
- The current implementation is broken. It include helm labels.
- There is no service for the webhook included

Fixes #

**Changes**

*
*

**Checklist:**

### DIFF
--- a/stable/vpa/templates/admission-controller-service.yaml
+++ b/stable/vpa/templates/admission-controller-service.yaml
@@ -2,18 +2,6 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ include "vpa.fullname" . }}-admission-controller
-spec:
-  ports:
-    - port: 443
-      targetPort: 8000
-  selector:
-    app.kubernetes.io/component: admission-controller
-    {{- include "vpa.selectorLabels" . | nindent 4 }}
----
-apiVersion: v1
-kind: Service
-metadata:
   name: vpa-webhook
 spec:
   ports:
--- a/stable/vpa/templates/admission-controller-service.yaml
+++ b/stable/vpa/templates/admission-controller-service.yaml
@@ -9,5 +9,17 @@ spec:
       targetPort: 8000
   selector:
     app.kubernetes.io/component: admission-controller
-    {{- include "vpa.labels" . | nindent 4 }}
+    {{- include "vpa.selectorLabels" . | nindent 4 }}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: vpa-webhook
+spec:
+  ports:
+    - port: 443
+      targetPort: 8000
+  selector:
+    app.kubernetes.io/component: admission-controller
+    {{- include "vpa.selectorLabels" . | nindent 4 }}
 {{- end }}
--- a/stable/vpa/Chart.yaml
+++ b/stable/vpa/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: vpa
 description: A Helm chart for Kubernetes Vertical Pod Autoscaler
 type: application
-version: 0.1.1
+version: 0.1.2
 appVersion: 0.8.0
 maintainers:
   - name: sudermanjr


stable/vpa/templates/admission-controller-service.yaml
```diff
@@ -9,5 +9,17 @@ spec:
...
+kind: Service
+metadata:
+  name: vpa-webhook
+spec:
+  ports:
+    - port: 443
+      targetPort: 8000
+  selector:
+    app.kubernetes.io/component: admission-controller
+    {{- include "vpa.selectorLabels" . | nindent 4 }}
```
sudermanjr: Perhaps I'm missing something, but this looks like the exact same service as above, but with a different name. Can you elaborate on why this is needed?

stable/vpa/templates/admission-controller-service.yaml
```diff
@@ -9,5 +9,17 @@ spec:
       targetPort: 8000
   selector:
     app.kubernetes.io/component: admission-controller
-    {{- include "vpa.labels" . | nindent 4 }}
+    {{- include "vpa.selectorLabels" . | nindent 4 }}
```
sudermanjr: Nice catch.

sudermanjr (changes requested):

stable/vpa/templates/admission-controller-service.yaml
dza89: The admission-controller pod creates a webhook, in this webhook is a service defined:

```kind: MutatingWebhookConfiguration
metadata:
  creationTimestamp: "2020-09-28T15:27:04Z"
  generation: 1
  name: vpa-webhook-config
  resourceVersion: "141705845"
  selfLink: /apis/admissionregistration.k8s.io/v1/mutatingwebhookconfigurations/vpa-webhook-config
  uid: 133e4c3b-0214-4b38-933b-69d60e290779
webhooks:
- admissionReviewVersions:
  service:
      name: vpa-webhook
      namespace: cpt
      port: 443
```

See also here below:
https://github.com/kubernetes/autoscaler/blob/8cd8009947a06949b9cd12c5a0d27f617c03e155/vertical-pod-autoscaler/deploy/admission-controller-deployment.yaml

The other one is for prometheus metrics i suppose

stable/vpa/templates/admission-controller-service.yaml
dza89: Yeah it included helm labels now

stable/vpa/templates/admission-controller-service.yaml
sudermanjr: I get that, but your PR creates two duplicate services to the same port on the same dpeloyment.

Even the example you posted only shows a single service. Can you explain why we need  two of them?

dza89: I think it's for metrics?
There also both in here
https://github.com/kubernetes/autoscaler/tree/8cd8009947a06949b9cd12c5a0d27f617c03e155/vertical-pod-autoscaler/deploy

dza89: You are right, thanks!
So remove the other one?

sudermanjr: I think just remove the duplicate one (the one you added), leave the first one, and make this PR just about fixing the label selector for the service.

dza89: Ok, but i still changed the naming as it won't work with the old name (referencing a non existing service in the webhook created)

stable/vpa/templates/admission-controller-service.yaml
```diff
@@ -2,12 +2,12 @@
...
-  name: {{ include "vpa.fullname" . }}-admission-controller
+  name: vpa-webhook
 spec:
   ports:
     - port: 443
       targetPort: 8000
   selector:
     app.kubernetes.io/component: admission-controller
-    {{- include "vpa.labels" . | nindent 4 }}
+    {{- include "vpa.selectorLabels" . | nindent 4 }}
```
sudermanjr, line 12: I just noticed that this will currently select the other pods from the recommender and the updater. We need to add in the `app.kubernetes.io/component: admission-controller` label as well.

```suggestion
    app.kubernetes.io/component: admission-controller
    {{- include "vpa.selectorLabels" . | nindent 4 }}
```

sudermanjr (changes requested):

sudermanjr (approved):

PR #369, opened 2020-10-01
add priorityclass

dza89:
**Why This PR?**
Add priorityclass option for all pods

Fixes #

**Changes**

Priorityclass added

**Checklist:**

### DIFF
--- a/stable/vpa/templates/admission-controller-deployment.yaml
+++ b/stable/vpa/templates/admission-controller-deployment.yaml
@@ -26,6 +26,10 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with .Values.priorityClassName }}
+      priorityClassName:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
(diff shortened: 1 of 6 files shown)


dza89:
Need some help with the check-docs, thought I fixed it but still error

sudermanjr (changes requested):

stable/vpa/templates/admission-controller-deployment.yaml
```diff
@@ -26,6 +26,9 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with .Values.priorityClassName }}
+      priorityClassName:
```
sudermanjr: I think you'll need to add the value too:

```suggestion
      priorityClassName: {{ . }}
```

sudermanjr (changes requested):
One more version bump after pulling should take care of the helm-docs and helm-lint.

~Can you also add `priorityClassName: testPriority` to the ci/test-values.yaml file? This will make sure this value gets end-to-end tested.~ - Don't do that, it will fail the CI

sudermanjr (approved):
Thanks!

PR #370, opened 2020-10-01
[insights-admission] configuration updates

rbren:
**Why This PR?**
Various changes after playing with the admission controller for a bit

Major fix for flakiness in the CA provider by removing `cert-manager.io/inject-apiserver-ca: "true"` - about 50% of requests were failing due to certs.

Fixes #

**Changes**
* helm hooks were messing with the uninstall, and seem to be unnecessary
* templates out some of the more important pieces of the config
* removes an (unnecessary?) CA provider
* sets default failurePolicy to Fail

**Checklist:**

### DIFF
--- a/stable/insights-admission/templates/cert.yaml
+++ b/stable/insights-admission/templates/cert.yaml
@@ -6,9 +6,8 @@ metadata:
 spec:
   commonName: {{ include "insights-admission.fullname" .}}.{{ .Release.Namespace}}.svc
   dnsNames:
-  - {{ include "insights-admission.fullname" . }}.{{ .Release.Namespace }}
   - {{ include "insights-admission.fullname" . }}.{{ .Release.Namespace }}.svc
-  - {{ include "insights-admission.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local
+  - {{ include "insights-admission.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
   issuerRef:
     kind: Issuer
     name: {{ include "insights-admission.fullname" . }}-selfsigned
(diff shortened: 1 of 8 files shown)


stable/insights-admission/templates/cert.yaml
```diff
@@ -3,52 +3,21 @@ apiVersion: cert-manager.io/v1alpha2
 kind: Certificate
 metadata:
   name: {{ include "insights-admission.fullname" . }}-cert
-  annotations:
-    "helm.sh/hook": pre-install
```
makoscafee, line 7: the need for this to be installed before any resources was removed right?

stable/insights-admission/templates/cert.yaml
```diff
@@ -3,52 +3,21 @@ apiVersion: cert-manager.io/v1alpha2
...
-    "helm.sh/hook-weight": "6"
 spec:
   commonName: {{ include "insights-admission.fullname" .}}.{{ .Release.Namespace}}.svc
   dnsNames:
+  - {{ include "insights-admission.fullname" . }}.{{ .Release.Namespace }}
   - {{ include "insights-admission.fullname" . }}.{{ .Release.Namespace }}.svc
   - {{ include "insights-admission.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local
```
makoscafee: it will be nice to refactor `cluster.local` to values. I have seen couple cluster change this to something else

makoscafee (approved):
LGTM. Few suggestions

stable/insights-admission/templates/cert.yaml
rbren: I'm not quite sure why it was here - my gut is that Bader was trying to control the order that these certs were installed.

Unfortunately, if you use `hook`, helm won't uninstall those resources when running `helm uninstall`, so I think it's best to remove it where possible. This installs and runs fine for me without it.

Will definitely want to get Bader's eyes on this when he comes back though to double check.

stable/insights-admission/templates/cert.yaml
rbren: Ooh good to know, thanks!

PR #371, opened 2020-10-01
Update helm-docs in the CI

sudermanjr:
**Why This PR?**
Users are reporting differences in newer version, so I'm bumping to the latest

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,8 +36,8 @@ jobs:
           name: Install helm-docs
           command: |
             cd /tmp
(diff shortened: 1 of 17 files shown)


rbren (approved):

rbren:
Thanks!

rbren:
Would be really cool if we could auto-run `helm-docs` using a GH action

sudermanjr:
> Would be really cool if we could auto-run `helm-docs` using a GH action

CI that makes changes to code is bad, mmmkay.

sudermanjr:
> Would be really cool if we could auto-run `helm-docs` using a GH action

I highly recommend using pre-commit. The file in the repo will auto-run helm-docs for you if you didn't and block your commit.

PR #373, opened 2020-10-05
[polaris] Allow to disable rbac and serviceAccount.

desaintmartin:
Allow to disable rbac and serviceAccount.

Some may want to setup by other means so that deploying this chart does not requires extended cluster-wide privileges.

Also update ClusterRole and ClusterRolesBinding to api v1.

**Checklist:**


rbren (approved):
Thanks for the contribution!

PR #374, opened 2020-10-07
update helm-docs to 1.3.0, sort based on file order

rbren:
**Why This PR?**
Updates `helm-docs` to latest, and uses the `--sort-values-order=file` argument to sort the value docs based on the order they appear in values.yaml

This gives us more control over how the values appear in the README.

I also updated the values order for Gemini and insights-admission - happy to rearrange others.

**Checklist:**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,14 +37,14 @@ jobs:
           name: Install helm-docs
           command: |
             cd /tmp
-            curl -LO https://github.com/norwoodj/helm-docs/releases/download/v0.15.0/helm-docs_0.15.0_Linux_x86_64.tar.gz
-            tar -zxvf helm-docs_0.15.0_Linux_x86_64.tar.gz
+            curl -LO https://github.com/norwoodj/helm-docs/releases/download/v1.3.0/helm-docs_1.3.0_Linux_x86_64.tar.gz
+            tar -zxvf helm-docs_1.3.0_Linux_x86_64.tar.gz
             mv helm-docs /usr/local/bin/helm-docs
             chmod +x /usr/local/bin/helm-docs
       - run:
           name: Check Docs for Changes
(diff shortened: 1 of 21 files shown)


baderbuddy (approved):
This looks good to me. It'll be nice having that control over the values.

makoscafee (approved):
LGTM

robertlucido (approved):
LGTMoi

sudermanjr:
Would this mean that people making PRs to the repo would have to pass that flag when making them?

sudermanjr (changes requested):
Would this mean that people making PRs to the repo would have to pass that flag when making them?

This also needs a documentation update in the Contributing.

If we are going to make a sweeping change like this, I would prefer doing the update on all the charts, not just yours

rbren:
Good catch! Updated CONTRIBUTING.

This does affect all charts using helm-docs. Like I said, happy to rearrange other values files if there are any that look particularly off, but I don't think I have quite enough context into each project to know what the best order is.

Maybe the best thing would be for chart owners to push any values.yaml rearrangements they'd like to see to this branch? Also happy to just make a best effort.

sudermanjr (changes requested):
Ah, I see you actually already did what I wanted, which was just running it on all the charts. That's fine for me, the codeowners can rearrange if they want.

The only other ask I have is to update the .pre-commit-config.yaml to use this flag so that those using pre-commit won't get bit by the change.


```
  - repo: https://github.com/norwoodj/helm-docs
    rev: v1.3.0
    hooks:
      - id: helm-docs
        args:
          - --sort-values-order=file
```

rbren:
👍 added pre-commit

sudermanjr (approved):
Just needs a rebase and I think it's probably good to go

PR #375, opened 2020-10-07
[Insights Agent] Add test of the chart.

baderbuddy:
**Why This PR?**
Adds an automated test to the Insights Agent chart. The new values are not documented in the README because they only apply for CI/CD

Fixes #

**Changes**

* Adds a new deployment and test pod for automated testing.
*

**Checklist:**


sudermanjr:
Why not just use helm test hooks?

baderbuddy:
Oh you had me doubting myself here :) I did on the Pod. The agent expects to be talking to Insights though along the way (from install instead of from when the tests start). So that's why there's a separate deployment for testing to mimic Insights.

rbren (approved):

stable/insights-agent/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to run the Fairwinds Insights agent
 name: insights-agent
-version: 0.27.5
+version: 0.27.6
 appVersion: 1.12.1
```
makoscafee: I think we are at `1.13.1` now

makoscafee (approved):
LGTM

PR #388, opened 2020-11-06
[Astro] Update README

baderbuddy:
**Why This PR?**
Updates the README to use helm-docs

Fixes #

**Changes**

* Adds documentation to Values.yaml and moves to using helm-docs

**Checklist:**

### DIFF
--- a/stable/astro/values.yaml
+++ b/stable/astro/values.yaml
@@ -1,7 +1,11 @@
 image:
+  # -- Docker image repo
   repository: quay.io/fairwinds/astro
+  # -- Docker image tag
   tag: "v1.5.3"
+  # -- Docker image pull policy
   pullPolicy: IfNotPresent
+# -- The resources block for the deployment.
 resources:
   limits:
     cpu: 100m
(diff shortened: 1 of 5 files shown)


sudermanjr (approved):
Approving my little portion

stable/astro/values.yaml
```diff
@@ -1,35 +1,56 @@
...
   serviceAccount:
+    # -- If true, a service account will be created. If false, you must set `deployment.serviceAccount.name`.
     create: true
-    # name: ExistingServiceAccountName
+    # -- The name of an existing service account to use.
+    name:
+  # -- The number of replicas to use.
+  replicas: 2
 secret:
+  # -- If true, a secret with API credentialss will be created. If false, you must set `secret.name`
```
lucasreed:
```suggestion
  # -- If true, a secret with API credentials will be created. If false, you must set `secret.name`
```

stable/astro/values.yaml
```diff
@@ -1,35 +1,56 @@
...
   create: true
-  # name: ExistingSecretName
+  # -- The name of an existing secret to mount to the container.
+  name:
```
lucasreed, line 42: Is it OK to leave this blank? previously it was commented out.

stable/astro/values.yaml
```diff
@@ -1,35 +1,56 @@
...
+  # -- If true, RBAC resources will be created.
   create: true
 deployment:
-  replicas: 2
   serviceAccount:
+    # -- If true, a service account will be created. If false, you must set `deployment.serviceAccount.name`.
     create: true
-    # name: ExistingServiceAccountName
+    # -- The name of an existing service account to use.
+    name:
```
lucasreed, line 35: Same question here about it being blank and previously being commented out. Is this ok?

lucasreed (commented):
One small spelling fix and a question about a particular value with no default that is not commented

stable/astro/values.yaml
baderbuddy: This is a weird syntax to me. But apparently that's one of the ways to represent nulls in Yaml. This way it still won't have a value.  This way we can still document it.

lucasreed (approved):

PR #393, opened 2020-11-17
Issue-391 - Create ServiceMonitor for prometheus scraping

### LINKED ISSUE #3: Create e2e tests using KIND for charts
Write a script to spin up a KIND cluster and install the chart(s) in order to test them.

### LINKED ISSUE #391: [rbac-manager] Add ServiceMonitor for prometheus scraping
I would like the helm chart for rbac-manager to include a headless service and prometheus operator servicemonitor to make the full integration experience with rbac-manager and monitoring smoother.  Currently we install rbac-manager via helm and then have to apply a service and servicemonitor manifest with the correct labels. As we don't control all of the labels used by the helm chart, any updates to these labels can result in needing to also update the service and servicemonitor.

I am happy to provide a PR with functionality if this is something that would be accepted.  I have found that the inclusion of these varies per OSS project (grafana, KIAM, and external-dns to name a few that do) while others choose not to as a ServiceMonitor is not a native k8s construct. 

Assuming this is something that would be approved are the following additions to the values.yaml acceptable:
```
monitoring:
  enabled: <true|false(default)>
  namespace: <"" | namespace> # defaults to "" / namespace the rbac-manager is installed into.  however some like to keep the servicemonitors in a different namespace>
  interval: 60s 
  # normally path and port are customizable here as well, but it appears you have the port and paths currently hard coded to 8080 and /metrics so I would leave those as they are.  
```

Please let me know.  Thanks,
Ryan

ryan-dyer-sp:
**Why This PR?**
Adds servicemonitor and headless service to allow for prometheus scraping
Fixes #391 

**Changes**
* New headless service and servicemonitor configurable through the .Values.serviceMonitor fields

**Checklist:**

### DIFF
--- a/stable/rbac-manager/values.yaml
+++ b/stable/rbac-manager/values.yaml
@@ -24,12 +24,8 @@ affinity: {}
 podAnnotations: {}
 podLabels: {}
 
-service:
-  enabled: false
-  annotations: {}
-
 serviceMonitor:
   enabled: false
-  annotations: ""
+  annotations: {}
   namespace: ""
   interval: 60s
--- a/stable/rbac-manager/values.yaml
+++ b/stable/rbac-manager/values.yaml
@@ -23,3 +23,13 @@ affinity: {}
 
 podAnnotations: {}
 podLabels: {}
+
+service:
+  enabled: false
+  annotations: {}
+
+serviceMonitor:
+  enabled: false
+  annotations: ""
+  namespace: ""
+  interval: 60s
--- a/stable/rbac-manager/Chart.yaml
+++ b/stable/rbac-manager/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: rbac-manager
-version: 1.6.1
+version: 1.6.2
 appVersion: 0.9.4
 description: A Kubernetes operator that simplifies the management of Role Bindings and Service Accounts.
 icon: https://raw.githubusercontent.com/FairwindsOps/charts/master/stable/rbac-manager/icon.png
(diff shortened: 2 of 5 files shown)


ryan-dyer-sp:
@sudermanjr 
Have looked at the failing insights build and I'm not seeing the issue as being related to these changes.  So could use some assistance in troubleshooting.  Thanks.

sudermanjr:
> Have looked at the failing insights build and I'm not seeing the issue as being related to these changes. So could use some assistance in troubleshooting. Thanks.

Don't worry about the insights check, that's non-blocking for us.

stable/rbac-manager/values.yaml
```diff
@@ -23,3 +23,13 @@ affinity: {}
...
 podAnnotations: {}
 podLabels: {}
+
+service:
+  enabled: false
+  annotations: {}
+
+serviceMonitor:
+  enabled: false
+  annotations: ""
```
sudermanjr: I know it doesn't change anything technically, but this will indicate to the end-user that this is a map instead of a string.

```suggestion
  annotations: {}
```

stable/rbac-manager/values.yaml
```diff
@@ -23,3 +23,13 @@ affinity: {}
 
 podAnnotations: {}
 podLabels: {}
+
+service:
+  enabled: false
+  annotations: {}
```
sudermanjr: It looks like these values aren't currently used. In the service template, you use the serviceMonitor values (which is what I was expecting)

```suggestion
```

stable/rbac-manager/values.yaml
sudermanjr: These annotations don't seem to get applied to the serviceMonitor, only the service. Is that intended? If so, maybe it should be a different key?

sudermanjr (changes requested):

sudermanjr (approved):
Looks great. Thanks!

PR #398, opened 2020-11-27
[gke-node-termination-handler] add .Values.extraArgs

mxmCherry:
**Why This PR?**
_It should be possible to configure grace period for regular pods, as documented [here](https://github.com/GoogleCloudPlatform/k8s-node-termination-handler#graceful-terminations-for-regular-pods-non-system-pods)_

**Changes**

* Adds `systemPodGracePeriod` option to chart values.

**Checklist:**

### DIFF
--- a/stable/gke-node-termination-handler/values.yaml
+++ b/stable/gke-node-termination-handler/values.yaml
@@ -24,3 +24,6 @@ resources:
   requests:
     cpu: 150m
     memory: 30Mi
+
+# systemPodGracePeriod -- A period to wait for regular pods to terminate. Recommended values: 0s-14s for preemptible VMs, 0s-(regular-vm-timeout/2) for regular VMs.
+systemPodGracePeriod: 0s
--- a/stable/gke-node-termination-handler/values.yaml
+++ b/stable/gke-node-termination-handler/values.yaml
@@ -25,5 +25,12 @@ resources:
     cpu: 150m
     memory: 30Mi
 
-# systemPodGracePeriod -- A period to wait for regular pods to terminate. Recommended values: 0s-14s for preemptible VMs, 0s-(regular-vm-timeout/2) for regular VMs.
-systemPodGracePeriod: 0s
+# args -- Command arguments. Usually you don't need to override them.
+args:
+  - "-v=10"
+  - "--logtostderr"
+  - "--exclude-pods=$(POD_NAME):$(POD_NAMESPACE)"
+  - "--taint=cloud.google.com/impending-node-termination::NoSchedule"
+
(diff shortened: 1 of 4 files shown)


stable/gke-node-termination-handler/values.yaml
```diff
@@ -24,3 +24,6 @@ resources:
   requests:
     cpu: 150m
     memory: 30Mi
+
+# systemPodGracePeriod -- A period to wait for regular pods to terminate. Recommended values: 0s-14s for preemptible VMs, 0s-(regular-vm-timeout/2) for regular VMs.
+systemPodGracePeriod: 0s
```
mxmCherry: I'm not sure if this is the best way to name it - maybe it should be smth like `args.systemPodGracePeriod` or so. Any suggestions?

mxmCherry:
Regarding [failed ci/circleci: insights](https://app.circleci.com/pipelines/github/FairwindsOps/charts/1859/workflows/92e67eed-dcd2-4820-9bc8-17deed993453/jobs/8064): seems to be fine as I am not supposed to have a FAIRWINDS_TOKEN, right?

sudermanjr:
> Regarding [failed ci/circleci: insights](https://app.circleci.com/pipelines/github/FairwindsOps/charts/1859/workflows/92e67eed-dcd2-4820-9bc8-17deed993453/jobs/8064): seems to be fine as I am not supposed to have a FAIRWINDS_TOKEN, right?

Yes, you can ignore that check for now. Sorry about the confusion

stable/gke-node-termination-handler/values.yaml
sudermanjr: I think maybe just adding an `extraArgs` field and then using the `toYaml` function might be best here, rather than trying to hard-code every possible flag.

sudermanjr (changes requested):

stable/gke-node-termination-handler/values.yaml
```diff
@@ -24,3 +24,13 @@ resources:
   requests:
     cpu: 150m
     memory: 30Mi
+
+# args -- Command arguments. Usually you don't need to override them.
```
mxmCherry, line 28: Maybe remove helm-docs comment so it's not shown in README? I don't think anyone would ever use this.

sudermanjr: I think it's fine to leave them. I would rather over-document than under.

sudermanjr (changes requested):
One last thing, would you mind changing the minor version rather than the patch? This add additional functionality, rather than just fixing a bug.

Otherwise looks great.

sudermanjr (approved):

PR #399, opened 2020-11-27
[stable/vpa] Add probes and metrics ports

QuentinBisson:
**Why This PR?**

This PR describes how to configure probes for the vpa chart as finding the correct port and paths needs users of the chart to dig into the vertical pod autoscaler pod

Fixes #

**Changes**

* This PR adds probes to vpa components as well as describe the metrics port to allow scraping of the vpa component metrics

**Checklist:**


sudermanjr (changes requested):
LGTM. Thanks!

In the future, we'll probably want to make these probes configurable, but this seems fine to me.

Only change requested is that we bump a minor version rather than patch. Changing probes could cause undesirable results or unexpected results on upgrade that I feel are better indicated by a minor version bump.

QuentinBisson:
I updated the chart version to 0.2.0

sudermanjr (approved):
Awesome. Thanks!!

PR #402, opened 2020-12-07
[Insights Agent] Disable Service Mesh

baderbuddy:
**Why This PR?**

Updates RBAC API versions to be consistent. Adds an option to disable injection of Service Mesh sidecars because those don't work with CronJobs.

Fixes #

**Changes**

*
*

**Checklist:**

### DIFF
--- a/stable/insights-agent/values.yaml
+++ b/stable/insights-agent/values.yaml
@@ -1,4 +1,3 @@
-disableServiceMesh: true
 
 insights:
   host: https://insights.fairwinds.com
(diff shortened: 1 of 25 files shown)


makoscafee (approved):
LGTM. 
This looks good probably giving user ability to add custom annotation as they see fit? just in case there are other sidecar injections that they need to skip.

stable/insights-agent/values.yaml
```diff
@@ -1,3 +1,5 @@
+disableServiceMesh: true
```
rbren: can we move this down? maybe in to the `cronjobs` section?

rbren (approved):

stable/insights-agent/values.yaml
baderbuddy: I forgot about the cronjobs section :) yeah that should work.

PR #412, opened 2020-12-16
scan more charts with insights

sudermanjr:


fairwinds-insights.yaml
```diff
@@ -26,3 +33,15 @@ exemptions:
       - hostPIDSet
       - runAsRootAllowed
     reason: kube-bench needs special permissions to do its job
+  - filename: goldilocks/goldilocks/charts/metrics-server/*
```
rbren, line 36: we've got a fix for exemptions going out today, but no `*` patterns. That's a really good feature though - opened an issue to track

rbren (commented):
Thanks for doing this! Will try and tackle that `panic` soon

PR #419, opened 2020-12-22
Update helm-docs to 1.4.0. Run pre-commit on everything

sudermanjr:
**Why This PR?**
Just a cleanup. I noticed some weirdness in the CI around docs, so I'm making sure everything is up to date.


baderbuddy (approved):
From my little piece this looks good. Is it worth adding a tools-version to control the version of helm-docs for asdf?

sudermanjr:
> From my little piece this looks good. Is it worth adding a tools-version to control the version of helm-docs for asdf?

I'm not a huge fan of that since it's already pinned in pre-commit and we will probably forget to update.

PR #426, opened 2021-01-14
[VPA] Add environment variable values to certGen container

### LINKED ISSUE #425: [vpa] Support HTTPS proxy in certgen Job

Currently when `admissionController.enabled=true` is enabled, by default certificates will generated as a `pre-install`/`pre-upgrade` hook in [`admission-controller-certgen.yaml`](https://github.com/FairwindsOps/charts/blob/master/stable/vpa/templates/admission-controller-certgen.yaml).

The `{{ include "vpa.fullname" . }}-certgen` Job requires public network access to [install packages](https://github.com/FairwindsOps/charts/blob/7464dcd2e8a430771557e05dacd95bbcd46f7fec/stable/vpa/templates/admission-controller-certgen.yaml#L76). In environments where public access is not allowed and a _HTTPS proxy_ must be used the Job fails:

```
fetch http://dl-cdn.alpinelinux.org/alpine/v3.10/main/x86_64/APKINDEX.tar.gz
ERROR: http://dl-cdn.alpinelinux.org/alpine/v3.10/main: network error (check Internet connection and firewall)
fetch http://dl-cdn.alpinelinux.org/alpine/v3.10/community/x86_64/APKINDEX.tar.gz
ERROR: http://dl-cdn.alpinelinux.org/alpine/v3.10/community: network error (check Internet connection and firewall)
(1/1) Installing openssl (1.1.1i-r0)
ERROR: openssl-1.1.1i-r0: network error (check Internet connection and firewall)
1 error; 81 MiB in 39 packages
```

The ability to pass in environment variables and/or additional arguments to the `Job`'s container spec would allow `https_proxy=...` to be specified to overcome this.

sudermanjr:
**Why This PR?**

Fixes #425

**Checklist:**

### DIFF
--- a/stable/vpa/templates/admission-controller-certgen.yaml
+++ b/stable/vpa/templates/admission-controller-certgen.yaml
@@ -117,5 +117,10 @@ spec:
             # Clean up after we're done.
             echo "Deleting ${TMP_DIR}."
             rm -rf ${TMP_DIR}
+        env:
+          {{- range $key, $value := .Values.admissionController.certGen.env }}
+          - name: {{ $key }}
+            value: {{ $value | quote }}
+          {{- end }}
 {{- end }}
 {{- end }}
--- a/stable/vpa/Chart.yaml
+++ b/stable/vpa/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: vpa
 description: A Helm chart for Kubernetes Vertical Pod Autoscaler
 type: application
-version: 0.2.1
+version: 0.2.2
 appVersion: 0.8.0
 maintainers:
   - name: sudermanjr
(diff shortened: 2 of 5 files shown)


stable/vpa/templates/admission-controller-certgen.yaml
```diff
@@ -117,5 +117,10 @@ spec:
             # Clean up after we're done.
             echo "Deleting ${TMP_DIR}."
             rm -rf ${TMP_DIR}
+        env:
```
rbren, line 120: If there are no keys, will this still be OK?

rbren (commented):
Looks like maybe an errant file called `stable/?`

stable/vpa/templates/admission-controller-certgen.yaml
sudermanjr: Yeah, because the default values file specifies `{}`

rbren: Oh that's what I meant - if it's an empty obect, the yaml ends up looking like
```
foo: bar
env:
baz: quuz
```
Does yaml treat that as `null`?

rbren (approved):

stable/vpa/templates/admission-controller-certgen.yaml
sudermanjr: Seems to be the case. Just tested it.

PR #428, opened 2021-01-15
[rbac-manager] Move CRDs to Helm3's standard crds/ folder

### LINKED ISSUE #427: [rbac-manager] Creation of CRDs should be optional
The Helm documentation states, that [Helm 3 does not support upgrading CRDs](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#install-a-crd-declaration-before-using-the-resource).
Helm suggests either using the [special crds folder of Helm 3](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you) or [managing the CRDs outside the chart](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-2-separate-charts).
In either case it would be great if the chart supported optional creation of the CRDs.

Make the creation of CRDs optional (for backwards compatibility this could be enabled by default).
Also move the CRDs to the Helm3 `crds` folder so standard mechanisms like `helm install --skip-crds` work.

Would you accept a PR with these changes?

Jeinhaus:
**Why This PR?**
Right now there is no way for a user to manage CRDs outside of this chart.
The recommended Helm3 way using `--skip-crds` does not work, since the CRD are placed in the `templates/` folder.
Fixes #427 

**Changes**
* Move the CustomResource Definitions to the Helm3 standard folder `crds/`
* Remove labels from the CRD since templating is not allowed/possible

This follows Helm3's recommended way of installing CRDs.
See the official [Helm3 documentation](https://helm.sh/docs/topics/charts/#custom-resource-definitions-crds).

Note though, that these CRDs are never upgraded by Helm3 after the first installation.
To manage CRDs follow the [guidelines in the Helm docs](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#install-a-crd-declaration-before-using-the-resource).

**Checklist:**

  * [ ] I'm not entirely sure if this counts as a breaking change 

### DIFF
--- a/stable/rbac-manager/Chart.yaml
+++ b/stable/rbac-manager/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: rbac-manager
-version: 1.7.0
+version: 1.8.0
 appVersion: 0.10.0
 description: A Kubernetes operator that simplifies the management of Role Bindings and Service Accounts.
 icon: https://raw.githubusercontent.com/FairwindsOps/charts/master/stable/rbac-manager/icon.png
--- a/stable/rbac-manager/crds/customresourcedefinition.yaml
+++ b/stable/rbac-manager/crds/customresourcedefinition.yaml
@@ -2,11 +2,6 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: rbacdefinitions.rbacmanager.reactiveops.io
-  labels:
-    app: {{ template "rbac-manager.name" . }}
-    chart: {{ template "rbac-manager.chart" . }}
-    release: {{ .Release.Name }}
-    heritage: {{ .Release.Service }}
 spec:
   group: rbacmanager.reactiveops.io
   names:


sudermanjr:
Not sure why the builds didn't trigger. I pushed an empty commit and they seem to be running now

sudermanjr (approved):
LGTM. Thanks. @lucasreed do you have time to check as well? I don't believe this will be breaking for us, but want to make sure.

lucasreed:
When I run an upgrade to this chart version with the previous iteration of the chart installed before, the CRD gets removed and not re-installed. A fresh install of rbac-manager at this chart version works fine.

Also, according [the docs](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#some-caveats-and-explanations) upgrading/deleting the CRD is not supported using this format.

I'd be more inclined to accept a PR that puts a boolean value in the helm chart that determines if the CRD should be installed or not. That way we keep the CRD in our templates and can continue to upgrade it, but the use-case of opting out is available too.

Jeinhaus:
I think the rationale from the Helm team, that upgrading a CRD is changing the cluster's API and may lead to unintentional data loss is valid. This part from the [CRD documentation](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#some-caveats-and-explanations) also brings up the point of validating a chart against the cluster's API which is not possible when upgrading the CRD with the chart:
>Another important point to consider in the discussion around CRD support is how the rendering of templates is handled. One of the distinct disadvantages of the crd-install method used in Helm 2 was the inability to properly validate charts due to changing API availability (a CRD is actually adding another available API to your Kubernetes cluster). If a chart installed a CRD, helm no longer had a valid set of API versions to work against. This is also the reason behind removing templating support from CRDs. With the new crds method of CRD installation, we now ensure that helm has completely valid information about the current state of the cluster.
This [StackOverflow post](https://stackoverflow.com/a/60295868) summarizes these points pretty good.

But I can see that this is a breaking change and might be slightly inconvenient. I'll open another PR and you can choose which one to merge :smile:

sudermanjr:
Closing in favor of the other option

PR #431, opened 2021-01-20
Add the ability to pass config to nova via configmap

sudermanjr:
**Why This PR?**
Nova will soon support adding config via a file, which will support passing desired-versions. This will allow us to specify that when deploying the agent.

**Changes**

* Add a configmap to nova called RELEASE-NAME-nova-config
* Mount the configmap to /config/nova.yaml
* Change the cmd for nova to be /nova find --config=/config/nova.yaml

**Checklist:**

### DIFF
--- a/stable/insights-agent/templates/nova/configmap.yaml
+++ b/stable/insights-agent/templates/nova/configmap.yaml
@@ -1,0 +1,9 @@
+{{- if .Values.nova.enabled -}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "insights-agent.fullname" $ }}-nova-config
+data:
+  nova.yaml: |
+{{ .Values.nova.config | toYaml | indent 4 }}
+{{- end }}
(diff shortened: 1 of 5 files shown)


stable/insights-agent/templates/nova/configmap.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if .Values.nova.enabled -}}
+{{- $_ := set . "Label" "nova" }}
```
baderbuddy: These 2 lines shouldn't be needed. If I remember right those are used by the CronJob templates.

baderbuddy (approved):
This feels so much cooler than what I envisioned when we were talking. I love that it's actual yaml in the values file so someone could just overwrite one of those values if they want to.

stable/insights-agent/templates/nova/configmap.yaml
sudermanjr: I think the top one should still be there. We don't want to create the configmap if nova isn't enabled.

But the second one I believe you are correct

sudermanjr:
There's definitely a bug with this implementation. Seems the uploadContainer template specifies `volumes` for the whole pod, so I can't just tack it on there. 😢 

EDIT: Fixed

rbren (approved):

sudermanjr:
I have tested this version in a cluster and it works exactly as intended.

PR #432, opened 2021-01-25
Unpin goldilocks image version

sudermanjr:
**Why This PR?**
The goldilocks image version was pinned a major version behind latest.

**Changes**

* Unpin the goldilocks image version so that we don't get behind
* Update the CI to install goldilocks/vpa/metrics-server
* Update goldilocks sub-chart to 3.2.1

**Checklist:**


sudermanjr:
Insights failures are on mostly unrelated charts. Some sub charts that I cannot exclude too.

baderbuddy:
It looks like at some point the Goldilocks output format might have changed. So we'll have to fix Insights to support the new format. Also I can't figure out how this works. The CronJob for Goldilocks is here in the Insights Agent templates, so where's it getting a version?

sudermanjr:
> The CronJob for Goldilocks is here in the Insights Agent templates, so where's it getting a version?

It's because we are overloading the subchart values and the insights goldilocks job values under `goldilocks`.  Since `goldilocks` is a sub-chart, it seems to merge the goldilocks values under that key.

PR #442, opened 2021-02-09
Adding PodDisruptionBudget to Vertical Pod Autoscaler Chart

mdgreenwald:
**Why This PR?**

This change creates a PodDisruptionBudget for the two pods in the Vertical Pod Autoscaler chart which allows them to be disrupted. It was observed that cluster scale downs would fail without this change.

Fixes #

**Changes**

* vpa-updater and vpa-reccomender both get PodDisruptionBudget templates.
* vpa-updater and vpa-reccomender both get entries in the values.yaml to configure them.
* README.md gets a note about this in the table as well.

**Checklist:**

* [✔️] I have updated the chart version in `Chart.yaml` following Semantic Versioning.
* [✔️] Any new values are backwards compatible and/or have sensible default.
* [✔️] Any new values have been added to the README for the Chart, or helm-docs has been run for the charts that support it.

Signed-off-by: Matthew Greenwald <matthew@fairwinds.com>


rbren (approved):

sudermanjr (approved):
awesome. Thanks!

PR #449, opened 2021-02-16
Update insights-agent chart to allow for two installations side by side

katiekeel:
**Why This PR?**
If you try to install a new deployment of `insights-agent` alongside an existing one - as we do installing `fw-telemetry` for clients who already have `insights-agent` present in their cluster - it will fail, because both deployments want to modify the same CRD.

**Fixes**

We need to allow the option to not install the CRD on a deployment.

**Changes**

* Update the insights-agent OPA `values.yaml` to take an `installCRDs` option, with default of `true`.
* Modify the CRD logic in the insights-agent OPA templates to require the flag to be true to install the CRD.

**Checklist:**

### DIFF
--- a/stable/insights-agent/README.md
+++ b/stable/insights-agent/README.md
@@ -79,6 +79,7 @@ Parameter | Description | Default
 `trivy.namespaceBlacklist` | Specifies which namespaces to not scan, takes an array of namespaces for example: `--set trivy.namespaceBlacklist="{kube-system,default}"` | nil
 `opa.role` | Specifies which ClusterRole to grant the OPA agent access to | view
 `opa.additionalAccess` | Specifies additional access to grant the OPA agent. This should contain an array of objects with each having an array of apiGroups, an array of resources, and an array of verbs. Just like a RoleBinding. | null
+`opa.installCRDs` | Specifies whether to install the `customcheckinstances.insights.fairwinds.com` CRD. If you are installing the `insights-agent` chart twice you will want to set this flag to `false` on *one* of the installs, doesn't matter which. | true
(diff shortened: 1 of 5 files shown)


stable/insights-agent/README.md
```diff
@@ -79,6 +79,7 @@ Parameter | Description | Default
 `trivy.namespaceBlacklist` | Specifies which namespaces to not scan, takes an array of namespaces for example: `--set trivy.namespaceBlacklist="{kube-system,default}"` | nil
 `opa.role` | Specifies which ClusterRole to grant the OPA agent access to | view
 `opa.additionalAccess` | Specifies additional access to grant the OPA agent. This should contain an array of objects with each having an array of apiGroups, an array of resources, and an array of verbs. Just like a RoleBinding. | null
+`opa.installCRDs` | Specifies whether to install the `customcheckinstances.insights.fairwinds.com` CRD. If you are installing the `insights-agent` chart twice - i.e. once as `insights-agent` and once as `fw-telemetry` - you will want to set this flag to `false` on *one* of the installs, doesn't matter which. | true
```
lucasreed: I don't think we need as much specificity here around fw-telemetry since this is a publicly consumable chart.

rbren (approved):
Thanks! 🎉

lucasreed (changes requested):
One minor change requested in-line.

Also you'll need to increment the chart version again since pulling in the updated master branch.

makoscafee (approved):

lucasreed (approved):
Awesome, thank you! 🎉

PR #452, opened 2021-02-22
[insights-agent] Add the ability to specify a config url for nova

sudermanjr:
**Why This PR?**
Nova recently added the ability to specify a config url. You can enable this by settign the `nova.configLocation` to an http(s) url. This will ignore the configmap

**Checklist:**

### DIFF
--- a/stable/insights-agent/templates/nova/cronjob.yaml
+++ b/stable/insights-agent/templates/nova/cronjob.yaml
@@ -27,7 +27,7 @@ spec:
             command:
               - /nova
               - find
-              - --config=/config/nova.yaml
+              - --config={{ .Config.configLocation }}
             {{ include "security-context" . | indent 12 | trim }}
           {{ include "uploaderContainer" . | indent 10 | trim }}
           volumes:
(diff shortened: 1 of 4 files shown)


stable/insights-agent/templates/nova/cronjob.yaml
```diff
@@ -27,7 +27,7 @@ spec:
             command:
               - /nova
               - find
-              - --config=/config/nova.yaml
+              - --config={{ .Config.configLocation }}
```
rbren, line 30: `.Values`?

sudermanjr: According to line three of that template, `.Config` is set to `.Values.nova`

sudermanjr: so no?

makoscafee (approved):
LGTM

stable/insights-agent/templates/nova/cronjob.yaml
rbren: Oh thank you. I'm even the one that set that up 🤦

rbren (approved):

PR #454, opened 2021-02-25
[Fairwinds Insights] Add Insights chart to charts repo

baderbuddy:
**Why This PR?**
Makes the Fairwinds Insights chart public for use in a Self Hosted world. This feels like it will work to successfully install Fairwinds Insights and should be flexible enough for multiple different use cases.

Fixes #

**Changes**

*
*

**Checklist:**


stable/fairwinds-insights/README.md.gotmpl
```diff
@@ -0,0 +1,13 @@
+# Fairwinds Insights
+
```
rbren, line 2: Can we put an `alpha` notice here for now?
```
> The self-hosted version of Fairwinds Insights is currently in alpha. The documentation is incomplete, and it is subject to breaking changes.
```

stable/fairwinds-insights/ci/test-values.yaml
```diff
@@ -0,0 +1,52 @@
+image:
+  tag: 3.2.4
+
+ingress:
+  hostedZones:
+  - test.cluster.local
+
+options:
+  insightsSAASHost: "https://staging.insights.fairwinds.com"
```
rbren, line 9: remove `staging`?

stable/fairwinds-insights/ci/test-values.yaml
```diff
@@ -0,0 +1,52 @@
...
+  autoGenerateKeys: true
+
+postgresql:
+  sslMode: "disable"
+  password: "hardcodedpassword"
```
rbren: how about `changeme`?

rbren (approved):
Would love @sudermanjr's feedback too

stable/fairwinds-insights/ci/test-values.yaml
baderbuddy: This is the file for CI and I was hesitant to hit prod and build up that data. I don't know if that's overly paranoid though.

stable/fairwinds-insights/Chart.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v2
+appVersion: 3.2.4
+description: A Helm chart to run the Fairwinds Insights server
+name: fairwinds-insights
+version: 1.0.0
```
sudermanjr: If we're saying alpha, this should probably be 0. something

sudermanjr (changes requested):
Still looking over this, but noticed one thing

stable/fairwinds-insights/templates/_env.yaml
```diff
@@ -0,0 +1,125 @@
+{{ define "env" }}
```
sudermanjr, line 1: I totally get why you broke this out, that's a lot of Env here. 

I'm curious though, do all deployments (now and future) need all these environment variables? I would be worried about accidentally leaking things like password to workloads that don't need it.

stable/fairwinds-insights/templates/aggregate-cronjob.yaml
```diff
@@ -0,0 +1,39 @@
...
+              securityContext:
+                readOnlyRootFilesystem: true
+                allowPrivilegeEscalation: false
+                privileged: false
+                runAsNonRoot: true
+                runAsUser: 10324
+                capabilities:
+                  drop:
+                    - ALL
+  {{ end }}
```
sudermanjr: nit: extra space. Also maybe `{{-` to chomp that line.

stable/fairwinds-insights/templates/deployment-api.yaml
```diff
@@ -0,0 +1,83 @@
...
+    metadata:
+      labels:
+        {{- include "fairwinds-insights.selectorLabels" . | nindent 8 }}
+        app.kubernetes.io/component: api
+    spec:
+      {{- with .Values.image.pullSecret }}
+      imagePullSecrets:
+        - name: {{ . }}
+      {{- end }}
+      serviceAccountName: {{ include "fairwinds-insights.fullname" . }}-insights
```
sudermanjr, line 23: In general, my guess is that _someone_ is going to want to be able to configure their own service account and then pass that name in. Might be better to consider for a future PR though

stable/fairwinds-insights/templates/deployment-dashboard.yaml
```diff
@@ -0,0 +1,73 @@
...
+      {{- end }}
+      containers:
+        - name: fw-insights-dashboard
+          image: "{{ .Values.dashboardImage.repository }}:{{ .Values.image.tag }}"
+          imagePullPolicy: Always
+          command: ["/bin/sh"]
+          args:
+          - -c
+          - |
+            {{ with .Values.options.dashboardConfig -}}
```
sudermanjr, line 31: This also doesn't seem to be in the values file. It also seems like something that might make a better configmap, but I don't actually know what's going on here.

sudermanjr (approved):
A couple small non-blocking things. Overall, awesome chart. Nice work

stable/fairwinds-insights/README.md
```diff
@@ -0,0 +1,88 @@
...
+| aggregateCronjob.schedules | list | `[{"cron":"5 0/2 * * *","interval":"120m","name":"bi-hourly"}]` | CRON schedules for the Workload Metrics aggregation job. |
+| emailCronjob.resources | object | `{"limits":{"cpu":"500m","memory":"1024Mi"},"requests":{"cpu":"80m","memory":"128Mi"}}` | Resources for the Action Items email job. |
+| emailCronjob.schedules | list | `[{"cron":"0 16 * * 1","interval":"168h","name":"weekly-email"}]` | CRON schedules for the Action Items email job. |
+| deleteOldActionItemsCronjob.resources | object | `{"limits":{"cpu":"500m","memory":"1024Mi"},"requests":{"cpu":"80m","memory":"128Mi"}}` | Resources for the delete old Action Items job. |
+| deleteOldActionItemsCronjob.schedules | list | `[{"cron":"0 0 * * *","interval":"24h","name":"ai-cleanup"}]` | CRON schedules for the delete old Action Items job. |
+| service.port | int | `80` | Port to be used for the API and Dashboard services. |
+| sanitizedBranch | string | `nil` | Prefix to use on hostname. Generally not needed. |
+| ingressApi.annotations | list | `[]` | Annotations to add to the API ingress. |
+| ingress.enabled | bool | `false` | Install Ingress objects. |
```
sudermanjr: I wonder if we might want to consider separating these? Someone might want an ingress for the dashboard, but not the api for example...

stable/fairwinds-insights/ci/test-values.yaml
sudermanjr: seems like a good idea to me.

stable/fairwinds-insights/templates/deployment-api.yaml
rbren: +1 on both counts. Would be good to do this for the agent chart as well

stable/fairwinds-insights/templates/_env.yaml
rbren: That's a good question. `now` the answer is yes, for the most part. I could see it getting misused in the future though.

Maybe we should have a few different envs that can get included down the line

PR #458, opened 2021-03-02
Add nova tmp for pulling config file from http. Add flag for logLevel to debug

sudermanjr:
**Why This PR?**
When using `nova --config=https://<>` nova expects to be able to use `/tmp` to store the file. This creates an emptyDir volume for that purpose

Fixes #

**Changes**

* Also adds a value `nova.logLevel` that gets passed, defaults to 3.

**Checklist:**


baderbuddy (approved):
LGTM, probably worth adding the logLevel parameter to the README though

rbren (approved):

PR #462, opened 2021-03-02
Bug fix for vpa running outside kube-system namespace

joshuastern:
**Why This PR?**
The VPA updater is unable to read the `leases` object when running outside the kube-system namespace.

**Changes**

* Set the `NAMESPACE` env var on the updater deployment. See where this var is used [here](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/updater/main.go#L66)

**Checklist:**

* [X] Any new values have been added to the README for the Chart, or helm-docs has been run for the charts that support it.


sudermanjr (approved):
Cool. I don't normally run the updater so I didn't catch this. Thanks!

PR #463, opened 2021-03-10
[fairwinds-insights] Small tweaks for quickstart

rbren:
**Why This PR?**

Fixes #

**Changes**

*
*

**Checklist:**


stable/fairwinds-insights/values.yaml
```diff
@@ -188,30 +190,32 @@ service:
...
   # -- Hostnames to use for Ingress
-  hostedZones: []
+  hostedZones: [""]
   # -- Annotations to add to the API and Dashboard ingresses.
-  annotations: []
+  annotations:
+
+ingressApi:
+  # -- Install API Ingress object.
+  enabled: true
```
baderbuddy: Oh we should leave these disabled. It's in the standards for the charts repo.

baderbuddy (commented):
I like all of this information, but I feel like most of this might be better served in the docs repo rather than with the chart. That feels more in line with how we've been doing things.

baderbuddy (approved):
LGTM

PR #465, opened 2021-03-24
[Insights Agent] Retrieve Polaris config from Insights

baderbuddy:
**Why This PR?**
Retrieves the config for Polaris from Insights on startup.

Fixes #

**Changes**

*
*

**Checklist:**

### DIFF
--- a/stable/insights-agent/templates/_specs.yaml
+++ b/stable/insights-agent/templates/_specs.yaml
@@ -42,6 +42,54 @@ volumeMounts:
   mountPath: /output
 {{- end }}
 
+{{ define "get-config-container" }}
+initContainers:
+- name: insights-config-downloader
+  image: "{{ .Values.uploader.image.repository }}:{{ .Values.uploader.image.tag }}"
+  imagePullPolicy: Always
+  command:
+  - ./download.sh
+  - --datatype
+  - {{ .Label }}
+  - --organization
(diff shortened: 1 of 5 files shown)


stable/insights-agent/templates/_specs.yaml
```diff
@@ -42,6 +42,54 @@ volumeMounts:
   mountPath: /output
 {{- end }}
 
+{{ define "get-config-container" }}
+initContainers:
```
rbren, line 46: Sweet, I wouldn't have thought to factor this out into something reusable!

stable/insights-agent/templates/polaris/cronjob.yaml
```diff
@@ -26,15 +27,20 @@ spec:
...
+            - |
+              CONFIG=""
+              if [ -f '/output/polaris-config.yaml' ]
+              then
+                CONFIG='--config /output/polaris-config.yaml'
+              fi
+              if [ -f '/opt/app/config.yaml' ]
+              then
+                CONFIG='--config /opt/app/config.yaml'
+              fi
```
rbren, line 42: So we always run the download script, and if it 404s we use the default config. Is that right?

rbren (approved):

makoscafee (approved):

PR #470, opened 2021-03-29
Add role parameter to letsencrypt-setup

dosullivan:
**Why This PR?**
Cert-manager allows for specifying an IAM role to assume for managing route53 records when using route53 to solve dns01 ACME challenges, see [here](https://cert-manager.io/docs/configuration/acme/dns01/route53/#creating-an-issuer-or-clusterissuer). 

This comes in handy when you're running a cluster in an AWS account separate from where your route53 records exist, as you can specify a role ARN in the account where the route53 zone lives. Cert-manager will assume it using its AWS credentials (as long as they have permission). I've tested this on EKS using EKS's service accounts, but it should work the same using kube2iam or kiam. 

There is actually a `role` parameter commented out in the values.yml example for the chart, but the template wasn't implemented yet. So, this is just adding that parameter to the template.

**Changes**

* Add a `role` parameter to the letsencrypt-setup chart template

**Checklist:**

* [ ✅ ] I have updated the chart version in `Chart.yaml` following Semantic Versioning.
* [ ✅ ] Any new values are backwards compatible and/or have sensible default.
* [ ✅  ] Any new values have been added to the README for the Chart, or helm-docs has been run for the charts that support it.

### DIFF
--- a/incubator/letsencrypt-setup/templates/_helpers.tpl
+++ b/incubator/letsencrypt-setup/templates/_helpers.tpl
@@ -60,6 +60,9 @@ solvers:
     route53:
       hostedZoneID: {{ .hostedZoneID | default "" }}
       region: {{ .region }}
+      {{- if .role }}
+      role: {{ .role }}
+      {{- end }} {{- /* endif .role */ -}}
       {{- if .secretName }}
       accessKeyID: {{ .accessKeyID }}
       secretAccessKeySecretRef:
--- a/incubator/letsencrypt-setup/Chart.yaml
+++ b/incubator/letsencrypt-setup/Chart.yaml
@@ -1,6 +1,6 @@
 name: letsencrypt-setup
 apiVersion: v1
-version: v2.2.1
+version: v2.2.2
 appVersion: v0.11.1
 description: Setup clusterIssuers for cert-manager
 maintainers:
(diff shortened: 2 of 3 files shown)


incubator/letsencrypt-setup/templates/_helpers.tpl
```diff
@@ -60,6 +60,9 @@ solvers:
     route53:
       hostedZoneID: {{ .hostedZoneID | default "" }}
       region: {{ .region }}
+      {{- if .role }}
+      role: {{ .role }}
+      {{- end }} {{- /* endif .role */ -}}
```
sudermanjr, line 65: Is the second thing just a comment? I haven't seen this pattern before, and it seems a bit unnecessary to me

dosullivan: Ha I hadn't either, I was just matching the pattern used in the rest of the file. It does seem pretty verbose.

sudermanjr: Huh, TIL. Works for me. We're probably going to be abandoning this chart in the near-ish future anyway.

sudermanjr (approved):

PR #481, opened 2021-05-05
[vpa] Add automountServiceAccountToken to vpa certGen and cleanup

joshuastern:
**Why This PR?**

Add ability to specify automountServiceAccountToken on the vpa certGen and cleanup service accounts

Fixes #

**Changes**

* Apply serviceAccount.automountServiceAccountToken value to certGen and cleanup service accounts
*

**Checklist:**

* [X] Any new values have been added to the README for the Chart, or helm-docs has been run for the charts that support it.

### DIFF
--- a/stable/vpa/Chart.yaml
+++ b/stable/vpa/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: vpa
 description: A Helm chart for Kubernetes Vertical Pod Autoscaler
 type: application
-version: 0.3.2
+version: 0.3.3
 appVersion: 0.9.2
 maintainers:
   - name: sudermanjr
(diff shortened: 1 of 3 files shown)


sudermanjr (changes requested):
I'm hesitant to add the `if` block around these particular service accounts using that same value. That value is intended to control the long-lived service accounts for the hooks.

I do understand that some people will want to disable these service accounts, but I don't think overloading that value is the right way to do that.

If you can elaborate on your use-case, maybe we can come up with a different plan

joshuastern:
> I'm hesitant to add the `if` block around these particular service accounts using that same value. That value is intended to control the long-lived service accounts for the hooks.
> 
> I do understand that some people will want to disable these service accounts, but I don't think overloading that value is the right way to do that.
> 
> If you can elaborate on your use-case, maybe we can come up with a different plan

I really only care about the `automountServiceAccountToken` flag. I will remove the `if` conditional block.

sudermanjr (approved):

j-land:
@sudermanjr ???

PR #490, opened 2021-06-02
[vault-token-injector] Allow setting vaultTokenFile

sudermanjr:
Blocked by the PR on vault-token-injector that creates these features. Expecting release 0.2.0

**Why This PR?**
Utilize a new feature to read the vault token from a file.

**Changes**
* Adds a new value to specify the location of the vault token file and pass that to an arg
* Add log verbosity
* Move circle token to a secret rather than just a printed env var

**Checklist:**


lucasreed (commented):
Wouldn't we need a configmap/secret to be created to mount as a file for this to work?

My brain is a bit tired so I could be misunderstanding.

sudermanjr:
I should have put a description:

This is intended to work with something like vault-k8s, which injects the token as a file into the running pod.

incubator/vault-token-injector/templates/deployment.yaml
```diff
@@ -37,14 +37,22 @@ spec:
...
+            - -v{{ .Values.logLevel }}
           volumeMounts:
           - name: config
             mountPath: /config
+          envFrom:
+          - secretRef:
+              name: {{ include "vault-token-injector.fullname" . }}
           env:
           - name: "CIRCLE_CI_TOKEN"
             value: {{ .Values.circleToken | quote }}
```
lucasreed: Should we remove this since we're pulling it from a secret now?

lucasreed (commented):
In-line question

incubator/vault-token-injector/templates/deployment.yaml
sudermanjr: Good catch

PR #495, opened 2021-06-09
[stackdriver-metrics-adapter] allow serviceAccount to have annotations set

lucasreed:


incubator/stackdriver-metrics-adapter/values.yaml
```diff
@@ -8,8 +8,14 @@ image:
...
 
+
+serviceAccount:
+# serviceAccount.annotations is a map of annotations you'd like added to the service account
+  annotations:
+#   key: value
+
 service:
-  type: ClusterIP
+  type: ClusterIPP
```
rbren: Is this right?

rbren (approved):

incubator/stackdriver-metrics-adapter/values.yaml
lucasreed: lol nope. I'll fix that.

PR #505, opened 2021-07-01
[insights-agent] Update kube-bench to 0.2, fix CronJob DaemonSet mode

ivanfetch:
* Update the Insights kube-bench plugin to version 0.2.0, which also
updates the kube-bench binary.
* The kube-bench CronJob in DaemonSet mode was failing to install with `Error: YAML parse error on insights-agent/templates/kube-bench/cronjob.yaml: error converting YAML to JSON: yaml: line 39: mapping values are not allowed in this context`
  * The `resources` block is now correctly indented for the DaemonSet mode of the kube-bench CronJob.
  * The `output` volume is now defined for the DaemonSet portion of the CronJob template.

**Checklist:**


rbren (approved):
Thanks Ivan!

ivanfetch:
I changed the chart version to a minor patch so the kube-bench update won't auto-release.

PR #507, opened 2021-07-01
[Astro] allow passing env variables to deployment

lucasreed:
**Why This PR?**
This allows passing env variables to the Astro deployment

Fixes #
In order to create monitors in an EU Datadog account, we need to allow
passing the following env var:

```yaml
env:
  name: DATADOG_HOST
  value: https://api.datadoghq.eu
```

**Changes**

* new `deployment.env` setting in the values file. Should be set as a
  map of key,value pairs
* Updated chart version

**Checklist:**

### DIFF
--- a/stable/astro/Chart.yaml
+++ b/stable/astro/Chart.yaml
@@ -1,5 +1,5 @@
 name: astro
-version: 1.0.7
+version: 1.0.8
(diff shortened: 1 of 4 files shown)


baderbuddy (approved):
LGTM

PR #508, opened 2021-07-06
Remove default runAsUser

Fodoj:
**Why This PR?**

Same as in https://github.com/FairwindsOps/charts/pull/271/files - fixed runAsUser breaks deployments on OpenShift. Alternative, I can adjust the code to allow skipping security context completely.

### DIFF
--- a/stable/goldilocks/templates/controller-deployment.yaml
+++ b/stable/goldilocks/templates/controller-deployment.yaml
@@ -41,8 +41,10 @@ spec:
             {{- range $name, $value := .Values.controller.flags }}
             - --{{ $name }}={{ $value }}
             {{- end }}
+        {{- if .Values.controller.securityContext }}
           securityContext:
             {{- toYaml .Values.controller.securityContext | nindent 12 }}
+        {{- end }}
           resources:
             {{- toYaml .Values.controller.resources | nindent 12 }}
       {{- with .Values.controller.nodeSelector }}
--- a/stable/goldilocks/Chart.yaml
+++ b/stable/goldilocks/Chart.yaml
@@ -4,7 +4,7 @@ description: |
   A Helm chart for running Fairwinds Goldilocks. See https://github.com/FairwindsOps/goldilocks
 name: goldilocks
 icon: https://raw.githubusercontent.com/FairwindsOps/charts/master/stable/goldilocks/icon.png
-version: 3.2.1
+version: 3.2.2
 maintainers:
   - name: sudermanjr
   - name: lucasreed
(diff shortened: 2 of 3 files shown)


sudermanjr:
You'll want to bump the chart version, rather than the `appVersion`.  App version dictates the image to pull. 

Additionally, I think it would be great if we could move to a completely configurable security context like you mentioned in the original description. That's the pattern I've started using on newer charts. That way we can leave the defaults as what they are now, but make the chart a lot more flexible going forward.

Fodoj:
@sudermanjr yep, I actually changed to allow completely skip SC for containers right after I created PR :) Updated, both implementation and adjusted the version.

Fodoj:
I can not access failed builds to see what exactly went wrong there, some hints appreciated

sudermanjr:
Don't worry about the insights ones. Can you not see the ones in Circle though?

sudermanjr:
```
[ERROR] templates/: parse error at (goldilocks/templates/dashboard-deployment.yaml:74): unexpected EOF
```

sudermanjr:
You can run `helm lint` on the chart to catch those before you push if that makes it easier.

Fodoj:
Login page of CircleCI confused me. Pushed the fix. 

I plan to add support for extraContainers for dashboard (so that it's possible to put oauth-proxy in front, for example) - is it ok to add externals dependencies like https://github.com/bitnami/charts/tree/master/bitnami/common to Fairwind Charts?

sudermanjr:
I would prefer to avoid adding any external sub-charts here since we can't control the changes that they make.

However, if it could be done in a way that is disabled by default, that might be okay

stable/goldilocks/templates/controller-deployment.yaml
```diff
@@ -41,8 +41,10 @@ spec:
             {{- range $name, $value := .Values.controller.flags }}
             - --{{ $name }}={{ $value }}
             {{- end }}
+        {{- if .Values.controller.securityContext }}
```
sudermanjr, line 44: Just realizing now that we already had the securityContext in here. Does setting `controller.securityContext: {}` not effectively do this in the existing version of the chart?

Fodoj: It doesn't - Helm will deep merge two hashes, so everything from the default values.yaml will still be there

sudermanjr: Interesting. I hadn't thought about that. 👍🏼

sudermanjr (approved):

PR #510, opened 2021-07-06
Add PDB for VPA admission-controller

### LINKED ISSUE #509: [vpa] PDB for admission controller
Draining of multiple nodes in parallel can cause the admission controller to become unavailable breaking the VPA flow.

Have an optional PDB in place for better resiliency

None

**Additional context**
The PDBs are already present for the recommender and updater components of the VPA

mikhailadvani:
**Why This PR?**

Adds PDB for vpa-admission-controller

Fixes #509 

**Changes**

* PDB for vpa-admission-controller with a default of maxUnavailable 1

**Checklist:**

### DIFF
--- a/stable/vpa/Chart.yaml
+++ b/stable/vpa/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: vpa
 description: A Helm chart for Kubernetes Vertical Pod Autoscaler
 type: application
-version: 0.3.4
+version: 0.3.5
 appVersion: 0.9.2
 maintainers:
   - name: sudermanjr
--- a/stable/vpa/README.md
+++ b/stable/vpa/README.md
@@ -95,6 +95,7 @@ recommender:
 | admissionController.certGen.env | object | `{}` | Additional environment variables to be added to the certgen container. Format is KEY: Value format |
 | admissionController.cleanupOnDelete | bool | `true` | If true, a post-delete job will remove the mutatingwebhookconfiguration and the tls secret for the admission controller |
 | admissionController.replicaCount | int | `1` |  |
+| admissionController.maxUnavailable | int | `1` | This is the max unavailable setting for the pod disruption budget |
 | admissionController.image.repository | string | `"us.gcr.io/k8s-artifacts-prod/autoscaling/vpa-admission-controller"` | The location of the vpa admission controller image |
 | admissionController.image.pullPolicy | string | `"Always"` | The pull policy for the admission controller image. Recommend not changing this |
 | admissionController.image.tag | string | `""` | Overrides the image tag whose default is the chart appVersion |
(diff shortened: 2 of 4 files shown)


sudermanjr:
Having the replica count set to 1 makes this a bit strange to add in my opinion. Does the vpa admission controller allow additional replicas? If so, then maybe we should change some defaults around here.

mikhailadvani:
agreed that replicaCount 1 & maxUnavailable 1 will be a blocker for node draining. I tried to align with the implementation of the recommender & updater. if a breaking change (values schema change) is agreeable, my recommendation would be to make the pod disruption configuration of maxUnavailable or minAvailable user configurable.
Samples:
- https://github.com/uswitch/kiam/pull/486/files
- https://github.com/aws/eks-charts/blob/master/stable/aws-load-balancer-controller/templates/pdb.yaml

mikhailadvani:
Have created an alternative solution in https://github.com/FairwindsOps/charts/pull/511 which addresses your concerns for all 3 components. Unfortunately, it is a breaking change in the chart

tyrken:
Just came here glad to find a PR already adding a PDB to the admission-controller (for reliability purposes, my current goal)!

From my understanding of the PDB spec replicaCount:1 and **maxUnavailable**:1 is fine, certainly this already exists by default for the updater & recommender, and I don't remember any issues draining nodes with them on in the past.  Hence would rather this proceed than make things more complex with #511

Note **minAvailable**:1 with the default replicaCount would be a blocker - are you misreading it is as that?

Related point - I see the two existing PDBs were explicitly added to **allow node draining to proceed** in #442 - which I find hard to understand given if absent they wouldn't stop node draining.  Maybe @mdgreenwald had an environment where PDBs were being added automatically to all deployments for safety reasons, and so adding one stopped a bad one somehow?

mikhailadvani:
@tyrken yes, my bad I read maxUnavailable and as minAvailable. agreed that it is not a blocker for node draining, but then again what is the value of a pdb with replicaCount 1 and maxUnavailable 1. hence IMHO, #511 is better that it does not gave a false sense of security and deploys a pdb only when it will actually have an impact

tyrken:
@mikhailadvani yes, I like the if statement in #511 where it only creates the PDB if replicaCount > 1 so would support adding that to this PR, though there is a corner case of someone wanting maxUnavailable:0 with replicaCount:1.  This is [explicitly suggested as a valid choice](https://kubernetes.io/docs/tasks/run-application/configure-pdb/#think-about-how-your-application-reacts-to-disruptions) in the K8s docs, though more usually for stateful singletons which VPA isn't.

But a bit concerned about the other breaking changes in #511 - needing to specify the entire `podDisruptionBudget` compared to this very easy maxUnavailable default which "just works".  Also a bit worried that I still don't understand the reasoning behind the original #442 PR.

sudermanjr:
I decided to go ahead with #511. That method seems much more effective, despite the small breaking change.

PR #511, opened 2021-07-06
Add PDB for VPA admission-controller

### LINKED ISSUE #509: [vpa] PDB for admission controller

mikhailadvani:
**Why This PR?**

Adds PDB for vpa-admission-controller. Makes the PDB configuration of recommender and updater more flexible.

Fixes #509. Addresses queries raised in #510 

**Changes**

* Have the ability to define PDB with minAvailable or maxUnavailable
* Create a PDB only if replicaCount is greater than 1. If the deployment is not running HA in the first place, we do not need a PDB.

**Checklist:**

### DIFF
--- a/stable/vpa/Chart.yaml
+++ b/stable/vpa/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: vpa
 description: A Helm chart for Kubernetes Vertical Pod Autoscaler
 type: application
-version: 0.3.4
+version: 0.4.0
 appVersion: 0.9.2
 maintainers:
   - name: sudermanjr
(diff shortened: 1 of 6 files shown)


sudermanjr:
This looks really great. I don't actually see where this is breaking for the chart. It seems 100% backward compatible to me, since you have the defaults in the values file.

Not sure what's going on with the CI. You might need to push an empty commit to kick it

mikhailadvani:
Backward compatible as long as the user take a look at the schema change in the values file

sudermanjr:
Ah, I see it now. 👍🏼

sudermanjr (approved):

PR #524, opened 2021-07-26
[fairwinds-insights] Better secrets management

rbren:
**Why This PR?**

Fixes #

**Changes**

*
*

**Checklist:**

### DIFF
--- a/stable/fairwinds-insights/templates/_env.yaml
+++ b/stable/fairwinds-insights/templates/_env.yaml
@@ -2,7 +2,7 @@
 # Required values
 envFrom:
 - secretRef:
-    name: fwinsights-secrets
+    name: {{ .Values.options.secretName }}
 env:
 - name: FAIRWINDS_HOST
   {{- if .Values.options.host }}
--- a/stable/fairwinds-insights/templates/_env.yaml
+++ b/stable/fairwinds-insights/templates/_env.yaml
@@ -33,6 +33,12 @@ env:
 - name: INSTALLATION_CODE
   value: {{ . | quote }}
 {{- end }}
+{{- with .Values.installationCodeSecret }}
+- name: INSTALLATION_CODE
+  secretKeyRef:
+    name: {{ printf "%s-%s" .Release.Name .Values.installationCodeSecret }}
+    key: INSTALLATION_CODE
+{{- end }}
 - name: CACHE_BUST_TOKEN
   value: {{ randAlphaNum 10 }}
 - name: FAIRWINDS_AGENT_CHART_TARGET_VERSION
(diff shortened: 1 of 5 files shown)


stable/fairwinds-insights/values.yaml
```diff
@@ -5,6 +5,9 @@ image:
 # -- Installation code provided by Fairwinds.
 installationCode:
 
+# -- Name of secret containing INSTALLATION_CODE
+installationCodeSecret:
```
makoscafee, line 9: this needs to be added in docs in README

stable/fairwinds-insights/values.yaml
```diff
@@ -39,6 +42,8 @@ options:
   autogenerateKeys: false
   # -- Run the job to migrate health scores to a new format
   migrateHealthScore: true
+  # -- Name of the secret where session keys and other secrets are stored
+  secretName: fwinsights-secrets
```
makoscafee, line 46: ditto ... on docs

stable/fairwinds-insights/templates/_env.yaml
```diff
@@ -33,6 +33,12 @@ env:
 - name: INSTALLATION_CODE
   value: {{ . | quote }}
 {{- end }}
+{{- with .Values.installationCodeSecret }}
+- name: INSTALLATION_CODE
+  secretKeyRef:
+    name: {{ printf "%s-%s" .Release.Name .Values.installationCodeSecret }}
```
makoscafee: mmh is there a way you are generating this secret? if is `.Values.installationCodeSecret` that the user supplies why we have release name here?

stable/fairwinds-insights/values.yaml
makoscafee: By the way you need to add `.values.SecretName` in secret https://github.com/FairwindsOps/charts/blob/master/stable/fairwinds-insights/templates/secret.yml#L10 it is still hardocoded one ... so even if new value is supplied here it will fail still.

stable/fairwinds-insights/values.yaml
```diff
@@ -5,6 +5,9 @@ image:
 # -- Installation code provided by Fairwinds.
 installationCode:
```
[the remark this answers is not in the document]
rbren: This is here for backwards compatibility

stable/fairwinds-insights/values.yaml
rbren: done!

stable/fairwinds-insights/values.yaml
rbren: Good catch!

stable/fairwinds-insights/templates/_env.yaml
rbren: 🤦 copypasta error. Good eyes!

makoscafee (approved):

PR #535, opened 2021-08-04
[goldilocks] Fix Ingress k8s API version

czunker:
**Why This PR?**

For k8s >= 1.19.0 it will create the Ingress object with networking.k8s.io/v1.
For version < 1.19.0 it will use networking.k8s.io/v1beta1.

**Changes**

**Checklist:**

Signed-off-by: Christian Zunker <christian.zunker@codecentric.cloud>


czunker:
I haven't really introduced new variables, but had to change the format for `paths`.
Before, it looked like this:
```
        paths:
          - /
```
Now, it has to look like this:
```
        paths:
          - path: /
            type: ImplementationSpecific
```
I think, the content of the `paths` list wasn't reflected in the docs, so I'm not quite sure how to handle this change. Any ideas?

stable/goldilocks/ci/ingress-values.yaml
```diff
@@ -0,0 +1,23 @@
+# For testing, make sure to install the vpa and metrics-server as sub-charts
```
sudermanjr: It doesn't seem necessary to add a new test file here. Please just modify the old one. What is added here doesn't conflict with that

stable/goldilocks/templates/dashboard-ingress.yaml
```diff
@@ -1,6 +1,12 @@
 {{- if and .Values.dashboard.enabled .Values.dashboard.ingress.enabled }}
 {{- $fullName := include "goldilocks.fullname" . -}}
+{{- $apiV1 := false -}}
+{{- if and (.Capabilities.APIVersions.Has "networking.k8s.io/v1") (semverCompare ">= v1.19.0" .Capabilities.KubeVersion.Version) -}}
```
sudermanjr, line 4: Awesome

stable/goldilocks/values.yaml
```diff
@@ -106,6 +106,8 @@ dashboard:
     hosts:
       - host: chart-example.local
         paths: []
+        # - path: /
+        #   type: ImplementationSpecific
```
sudermanjr: I don't think these need to be commented. In fact, I'm realizing this will actually be backward compatible if they are not

sudermanjr (changes requested):

czunker:
Thanks for your review. Should all be fixed now.

AlexanderBabel:
@czunker Hey, it would be great if you also add support for `ingressClassName`. This field replacing the annotation `kubernetes.io/ingress.class`

https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation

Thank you in advance!

czunker:
@AlexanderBabel Good point.
Because the field needs support from the installed ingress controller, it's up to the user to decide whether to use the filed or the annotation.

@sudermanjr Could you please review again?

czunker:
The CI creates a different `helm-docs` output to the one generated on my machine.

Perhaps different versions? I use
```
helm-docs --version
helm-docs version 1.5.0
```

lucasreed:
@czunker It looks like we are also using 1.5.0 in CI: https://github.com/FairwindsOps/charts/blob/master/.circleci/config.yml#L40

If you could rebase and fix the merge conflicts we'll get another review going, sorry for the delay!

czunker:
@lucasreed Thanks for the update. Fixed the merge conflicts.

lucasreed:
> @lucasreed Thanks for the update. Fixed the merge conflicts.

Thanks! Also, when you run helm-docs, are you using this command: `helm-docs --sort-values-order=file` ? That is what runs in CI to check for diffs.

czunker:
Thanks @lucasreed 
That fixed it.

stable/goldilocks/values.yaml
lucasreed: I agree, I think we can uncomment this and remove the `[]` from line 118

lucasreed (changes requested):

stable/goldilocks/values.yaml
czunker: Fixed

lucasreed (dismissed):

lucasreed:
Thanks for the contribution @czunker! LGTM, I'll let @sudermanjr look it over one more time as well.

stable/goldilocks/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 appVersion: "v3.1.4"
-version: 3.2.7
+version: 3.2.8
```
lucasreed: Because of the possibility that someone would have breaking changes here if they are deploying currently with an ingress path definition that does not include `type` I think we should make this a major release version change to `4.0.0`

lucasreed (changes requested):
Apologies for the back and forth here, but can you update the chart version (my inline comment) and also update the README with a note about upgrading to v4.0.0 that any ingress path definitions will require the `type` value to be defined now.

czunker:
> Apologies for the back and forth here, but can you update the chart version (my inline comment) and also update the README with a note about upgrading to v4.0.0 that any ingress path definitions will require the `type` value to be defined now.

No Problem.

Bump the version and added a change note. Please have a look.

lucasreed (approved):
Thanks again! LGTM

rust84:
Any chance that this will be merged soon?

AlexanderBabel:
@rust84 If you are using the `subcharts` pattern, then you can just disable the ingress from this chart and define your ingress resource inside the `templates` directory. That's my current solution until this PR gets finally merged.

sudermanjr:
Looks like we lost track of this. I updated the PR to resolve the conflicts. @lucasreed can you please double check this?

stable/goldilocks/Chart.yaml
```diff
@@ -22,4 +22,4 @@ dependencies:
 - name: metrics-server
   version: 5.3.3
   repository: https://charts.bitnami.com/bitnami
-  condition: metrics-server.enabled
+  condition: metrics-server.enabled
```
sudermanjr, line 25:
```suggestion
  condition: metrics-server.enabled

```

sudermanjr:
```suggestion
  condition: metrics-server.enabled

```

stable/goldilocks/ci/test-values.yaml
```diff
@@ -38,3 +45,4 @@ dashboard:
   deployment:
     additionalLabels:
       test: value
+
```
sudermanjr:
```suggestion
```

lucasreed (approved):

sudermanjr (approved):

PR #552, opened 2021-09-01
Add OpenAPI chart

makoscafee:
**Why This PR?**

Fixes #

**Changes**

*
*

**Checklist:**


rbren (approved):

stable/fairwinds-insights/templates/ingress.yaml
```diff
@@ -122,6 +126,10 @@ spec:
             {{ else }}
             path: /
             {{ end }}
+          - backend:
+              serviceName: {{ $fullName }}-open-api
+              servicePort: http
+            path: /swagger
```
rbren, line 139: Just need to support `starPaths` like above

makoscafee:
so, swagger update `usr/share/nginx/html/index.html` with the `SWAGGER_JSON_URL` just after the container has been  created. i have changed `readOnlyRootFilesystem` to default `false` not sure this is the best approach though.

stable/fairwinds-insights/templates/deployment-open-api.yaml
```diff
@@ -0,0 +1,54 @@
...
+        - name: fairwinds-insights
+          image: "{{ .Values.openApiImage.repository }}:{{ include "fairwinds-insights.openApiImageTag" . }}"
+          imagePullPolicy: Always
+          ports:
+          - name: http
+            containerPort: {{ .Values.openApi.port }}
+            protocol: TCP
+          env:
+          - name: SWAGGER_JSON_URL
+            value: {{ include "fairwinds-insights.fullname" . }}-api/v0/swagger.json
```
rbren, line 34: does this work? Just the service name?

rbren:
Hmm - I don't see another way around it. Thanks Barnie!

PR #565, opened 2021-09-20
[fairwinds-insights] Use ingress host instead

makoscafee:
**Why This PR?**

since swagger ui endpoint will be accessible over `https` using service name with http will have some cors issues... the best way is to use ingress host itself.
see
Screenshot 2021-09-20 at 23 36 54

Fixes #

**Changes**

*
*

**Checklist:**


rbren (approved):
🎉

PR #569, opened 2021-09-23
[vault-token-injector] Update vault token injector to v1.3.0

sudermanjr:
**Checklist:**


lucasreed (commented):
Should we also add the warning message here about the klog level when configured to `10`?

sudermanjr:
@lucasreed  I added a warning about that to the values file doc string

lucasreed (approved):

rbren (approved):

PR #587, opened 2021-10-25
[insights-admission] Add isCA: true

rbren:
**Why This PR?**

Fixes #

**Changes**

*
*

**Checklist:**

### DIFF
--- a/stable/insights-admission/templates/cert.yaml
+++ b/stable/insights-admission/templates/cert.yaml
@@ -14,6 +14,7 @@ spec:
   dnsNames:
   - {{ include "insights-admission.fullname" . }}.{{ .Release.Namespace }}.svc
   - {{ include "insights-admission.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
+  isCA: true
   issuerRef:
     kind: Issuer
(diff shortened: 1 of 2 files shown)


stable/insights-admission/templates/cert.yaml
```diff
@@ -14,6 +14,7 @@ spec:
   dnsNames:
   - {{ include "insights-admission.fullname" . }}.{{ .Release.Namespace }}.svc
   - {{ include "insights-admission.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
+  isCA: true
```
makoscafee, line 17: shouldn't we refactor this to `values.yaml` so people could be able to set it either true or false?

makoscafee (approved):

stable/insights-admission/templates/cert.yaml
rbren: Do you think there's a case where it should be `false`? Andy and I were thinking it should always be true, per [this article](https://cert-manager.io/docs/configuration/selfsigned/#bootstrapping-ca-issuers)

makoscafee: aah yes, I agree in this context it should always be `true`.

PR #593, opened 2021-10-27
[insights-agent] Exempt polaris check for prometheus sever

makoscafee:
**Why This PR?**

Fixes #

**Changes**

*
*

**Checklist:**

### DIFF
--- a/stable/insights-agent/values.yaml
+++ b/stable/insights-agent/values.yaml
@@ -331,7 +331,7 @@ prometheus:
       capabilities:
         drop:
           - ALL
-    podAnnotations: 
+    podAnnotations:
       polaris.fairwinds.com/insecureCapabilities-exempt: "true"
       polaris.fairwinds.com/privilegeEscalationAllowed-exempt: "true"
     resources:
--- a/stable/insights-agent/values.yaml
+++ b/stable/insights-agent/values.yaml
@@ -297,9 +297,6 @@ prometheus:
   kubeStateMetrics:
     enabled: true
     kube-state-metrics:
-      podAnnotations: 
-        polaris.fairwinds.com/insecureCapabilities-exempt: "true"
-        polaris.fairwinds.com/privilegeEscalationAllowed-exempt: "true"
       resources:
         limits:
           cpu: 100m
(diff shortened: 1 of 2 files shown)


stable/insights-agent/values.yaml
```diff
@@ -307,6 +320,28 @@ prometheus:
...
+      runAsNonRoot: true
+      runAsGroup: 65534
+      fsGroup: 65534
+      allowPrivilegeEscalation: false
+      capabilities:
+        drop:
+          - ALL
+    podAnnotations:
+      polaris.fairwinds.com/insecureCapabilities-exempt: "true"
+      polaris.fairwinds.com/privilegeEscalationAllowed-exempt: "true"
```
rbren, line 325: It looks like we can drop these based on the security context you're setting

rbren (commented):
Thanks Barnie!

Were you able to verify that the report still runs correctly?

stable/insights-agent/values.yaml
makoscafee: Apparently those are `podSecurityContext` and not `containerSecurityContext` so 
```
allowPrivilegeEscalation: false
        capabilities:
          drop:
            - ALL
```
does not work for podSecurityContext.

makoscafee: removed security context for `prometheus-server`

makoscafee:
and verified everthing is working -> https://staging.insights.fairwinds.com/v0/organizations/acme-co/clusters/umoja/reports/resource-metrics/detail?timestamp=2021-10-27T16:31:01.666

rbren:
@makoscafee can you install polaris and kubesec in that staging environment to see if the action items got fixed?

stable/insights-agent/values.yaml
```diff
@@ -296,6 +296,19 @@ resourcemetrics:
 prometheus:
   kubeStateMetrics:
     enabled: true
+    kube-state-metrics:
```
rbren: is this extra key intentional?

makoscafee: good catch ... those are values for `kube-state-metrics` moved them to parent `prometheus` chart

rbren (approved):

PR #601, opened 2021-11-03
[insights-agent] update dependencies

rbren:
**Why This PR?**

Fixes #

**Changes**

*
*

**Checklist:**

### DIFF
--- a/stable/insights-agent/requirements.lock
+++ b/stable/insights-agent/requirements.lock
@@ -4,9 +4,9 @@ dependencies:
   version: 4.0.1
 - name: prometheus
   repository: https://prometheus-community.github.io/helm-charts
-  version: 14.7.1
+  version: 14.11.0
 - name: insights-admission
   repository: https://charts.fairwinds.com/stable
   version: 0.4.0
-digest: sha256:fe736cf2c93b54f2f04f24c843dce3758d3f21b47e324c1962a5c2da1c8b97aa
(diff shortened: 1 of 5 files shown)


stable/insights-agent/requirements.lock
```diff
@@ -1,12 +1,12 @@
...
   repository: https://charts.fairwinds.com/stable
-  version: 3.2.7
+  version: 4.0.1
 - name: prometheus
   repository: https://prometheus-community.github.io/helm-charts
-  version: 14.7.1
+  version: 14.11.0
 - name: insights-admission
   repository: https://charts.fairwinds.com/stable
   version: 0.4.0
```
ivanfetch-wt, line 10: I'm not familiar with the "generated" aspect of this, but I wonder whether its version should change upstream, since it was re-generated and has a different hash? Feel free to ignore this comment if it is something I have yet to understand.

ivanfetch-wt (approved):

stable/insights-agent/requirements.lock
rbren: I think you're on the right track - it's changed in requirements.yaml, and then this file gets updated when you run `helm dependency update`

ivanfetch-wt: It looks like the `admission` version remains at 0.4.0, yet the `digest` and `generated` in release.lock changed?

PR #604, opened 2021-11-07
[goldilocks] enable to pull the uninstall vpa image from custom repos…

yasintahaerol:
…itory

**Why This PR?**
_a short description of why this PR is needed_
This image was pulling from a constant repository. Some developers or organizations want to use own their repository. It will be enabled to use custom repository by this feature. 
Fixes #
It will be enabled to use custom repository by this feature.

**Changes**
Changes proposed in this pull request:

*
*

**Checklist:**

* [*] I have updated the chart version in `Chart.yaml` following Semantic Versioning.
* [*] Any new values are backwards compatible and/or have sensible default.
* [*] Any new values have been added to the README for the Chart, or helm-docs has been run for the charts that support it.

PR #608, opened 2021-11-09
[goldilocks] enable to pull uninstall image from custom repository

yasintahaerol:
**Why This PR?**

This image was pulling from a constant repository. Some developers or organizations want to use own their repository. It will be enabled to use custom repository by this feature.

Fixes #

It will be enabled to use custom repository by this feature.

**Changes**

*
*

**Checklist:**

### DIFF
--- a/stable/goldilocks/Chart.yaml
+++ b/stable/goldilocks/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 appVersion: "v4.0.0"
-version: 4.0.1
+version: 4.0.2
 description: |
   A Helm chart for running Fairwinds Goldilocks. See https://github.com/FairwindsOps/goldilocks
 name: goldilocks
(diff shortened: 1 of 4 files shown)


sudermanjr:
This look like a duplicate of #604 which was closed. Am I missing something here?

yasintahaerol:
Yes, it is same. First time I used helm-docs and could not figure out how it works. Because of this reason I closed to old one and want to make fresh start. I hope everything is okey 🙏

sudermanjr:
@yasintahaerol We are actually planning to deprecate this feature (uninstallVPA) soon, especially since it is only needed to upgrade from very old versions of this chart. I don't think we should continue to update a feature that's going to be removed.

@lucasreed please correct me if I'm wrong.

yasintahaerol:
thaks, yes i aggree with you. I closed the PR.

PR #609, opened 2021-11-09
[insights-agent] Falco plugin

makoscafee:
**Why This PR?**

Fixes #

**Changes**

*
*

**Checklist:**


stable/insights-agent/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to run the Fairwinds Insights agent
 name: insights-agent
-version: 1.15.4
+version: 1.15.5
```
rbren: I think we can go to 16

stable/insights-agent/ci/test-values.yaml
```diff
@@ -59,3 +59,11 @@ awscosts:
   tagkey: kubernetes_cluster
   tagvalue: staging.internal.reactiveops.com
   workgroup: cur_athena_workgroup
+
+falcoagent:
+  enabled: true
+  image:
+    tag: "build_5220"
```
rbren: We should remove this before merging

stable/insights-agent/requirements.yaml
```diff
@@ -11,3 +11,7 @@ dependencies:
   repository: https://charts.fairwinds.com/stable
   version: 0.4.0
   condition: admission.enabled
+- name: falco
+  repository: https://falcosecurity.github.io/charts
+  version: 1.16.0
+  condition: falcoagent.installFalco
```
rbren: I think this should default to `true`

stable/insights-agent/templates/falco/deployment.yaml
```diff
@@ -0,0 +1,107 @@
...
+    {{- if .Values.falcoagent.priorityClassName }}
+      priorityClassName: {{ .Values.falcoagent.priorityClassName }}
+    {{- end }}
+      containers:
+      - name: falco-agent
+        volumeMounts:
+        - name: output-data
+          mountPath: /output
+      {{- if .Values.falcoagent.securityContext.enabled }}
+        securityContext:
```
rbren, line 62: Maybe do something like [this](https://github.com/FairwindsOps/charts/blob/43190a838474fb386d135fbf99012952ad3f6feb/stable/goldilocks/templates/controller-deployment.yaml#L56), where you just dump the whole security context

stable/insights-agent/values.yaml
```diff
@@ -388,3 +388,58 @@ awscosts:
     requests:
       cpu: 100m
       memory: 128Mi
+
+falcoagent:
```
rbren: I'd love to call this `falco` instead of `falcoagent`. I think the reason you went with `falcoagent` is so that you can override the values for the falco subchart independently of our agent - is that right?

If so, I think we can use [helm aliases](https://helm.sh/docs/topics/charts/#alias-field-in-dependencies) to put the subchart values in a section called `falcoOptions` or similar (not sure on what the best name would be...)

stable/insights-agent/values.yaml
```diff
@@ -388,3 +388,58 @@ awscosts:
...
+  enabled: false
+  installFalco: false
```
rbren, line 394: I think this should default to `true`

stable/insights-agent/values.yaml
```diff
@@ -388,3 +388,58 @@ awscosts:
...
+  statefulset: false
```
rbren: I think this should default to `true`

stable/insights-agent/values.yaml
```diff
@@ -388,3 +388,58 @@ awscosts:
...
+  persistence:
+    enabled: false
```
rbren: I think this should default to `true`

stable/insights-agent/values.yaml
```diff
@@ -388,3 +388,58 @@ awscosts:
...
+    storageClass: ""
+    accessModes:
+      - ReadWriteOnce
+    annotations: {}
+    ## If defined, PVC must be created manually before volume will be bound
+    existingClaim: ""
```
rbren: Did this come from gemini? I don't think it's being used here

stable/insights-agent/templates/falco/deployment.yaml
```diff
@@ -0,0 +1,107 @@
...
+      labels:
+        app: falco-agent
+    {{- if .Values.falcoagent.podAnnotations }}
+      annotations:
+        {{ toYaml .Values.falcoagent.podAnnotations | indent 8 }}
+    {{- end }}
+    spec:
+      serviceAccountName: {{ include "insights-agent.fullname" . }}-{{ .Label }}
+      {{- if .Values.falcoagent.securityContext.enabled }}
+      securityContext:
```
rbren, line 51: ditto on copying securityContext

stable/insights-agent/templates/falco/deployment.yaml
```diff
@@ -0,0 +1,107 @@
...
+  {{- include "metadata" . }}
+spec:
+  selector:
+    matchLabels:
+      app: falco-agent
+{{- if .Values.falcoagent.statefulset }}
+  serviceName: falco-agent
+{{- end }}
+{{- if .Values.falcoagent.persistence.enabled }}
+  volumeClaimTemplates:
```
rbren, line 20: I think this is only for statefulsets, right?

stable/insights-agent/ci/test-values.yaml
makoscafee: noted

stable/insights-agent/values.yaml
makoscafee: Actually here should be false... if true it will be installed regardless if falco.enabled = true

rbren: The way I'm reading [this](https://github.com/FairwindsOps/charts/pull/609/files#diff-b839fecfa2c0220c0538af8e900cf7c81599a5fbc2450f840a28e1a4f346d480R17), the subchart doesn't get installed unless this is `true` - do I have that wrong?

makoscafee: That is very correct ... my point is say this `installFalco` is by default `true` but `falcoagent.enabled` is `false` then subchart still gets installed.

makoscafee: while the agent/report is not enabled

rbren (approved):

PR #612, opened 2021-11-10
[goldilocks] Add support for VPA's including or not namespaces

varet80:
Extends Goldilocks chart, Add support for the controller flag `--on-by-default`

- Add `--on-by-default` flag support
- Enable also `--exclude-namespaces` when `--on-by-default` is enabled
- Support `--include-namespaces`, when `--on-by-default` is not enabled

**Changes**

* Added helm chart changes to support the above mentioned workflow
**Checklist:**


sudermanjr:
@billiaz it looks like multiple email addresses were used in the commits. Can you please add that email address to your github profile for the CLI signing?

RE: the PR. This seems fine. I'm generally not a huge fan of adding things that are enabled by extraArgs, but I'm not wholly opposed to merging this. @lucasreed would love your opinions on this.

lucasreed:
I think the main reason we have the `controller.flags` field here is so we don't need to support every possible flag with a helm chart value.

This issue is already solvable by adding something like this to your values when installing:
```
controller:
  flags:
    on-by-default: true
    exclude-namespaces: ns-1,ns-2
```

I think my vote currently is to leave this out and ask that the `controller.flags` value is utilized instead. Is there a good argument for merging this vs. using that value that already exists?

lucasreed:
Per my last comment, I think we'll close this as the functionality should already work with the proper `controller.flags` values.

PR #615, opened 2021-11-11
Add RightSizer to insights-agent

ivanfetch-wt:
This adds both a [RightSizer controller](https://github.com/FairwindsOps/insights-plugins/pull/530) and a cron job. Each have their own RBAC. Note the cron job does not use a typical "agent" container, because the Insights Uploader is able to pull report data from a ConfigMap that the RightSizer controller maintains.

**Checklist:**


stable/insights-agent/templates/right-sizer/controller-rbac.yaml
```diff
@@ -0,0 +1,110 @@
+{{- if and .Values.rightsizer.enabled (not .Values.rbac.disabled) -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "insights-agent.fullname" . }}-right-sizer-controller
+  labels:
+    app: insights-agent
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
rbren, line 10: we should just use the `view` clusterrole

stable/insights-agent/values.yaml
```diff
@@ -389,6 +389,18 @@ awscosts:
...
       memory: 128Mi
 
+rightsizer:
+  enabled: false
+  schedule: "rand * * * *"
+  timeout: 300
+  # This image is for the controller, the agent only runs the Insights uploader.
+  image:
+    repository: quay.io/fairwinds/right-sizer
+    tag: right-sizer-jam
```
makoscafee: we need to change this to actual release version

stable/insights-agent/templates/right-sizer/controller-deployment.yaml
```diff
@@ -0,0 +1,35 @@
...
+    matchLabels:
+      component: right-sizer
+      app: insights-agent
+  template:
+    metadata:
+      labels:
+        component: right-sizer
+        app: insights-agent
+    spec:
+      containers:
```
makoscafee, line 24: will probably need resources configuration here.

stable/insights-agent/templates/right-sizer/controller-deployment.yaml
```diff
@@ -0,0 +1,35 @@
...
+        - name: {{ .Chart.Name }}-right-sizer-controller
+          ports:
+          - containerPort: 8080
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
```
makoscafee: mmh do we have `Values.securityContext defined any where? and i would assume if they were to be defined should be in under `Values.rightsizer.securityContext`

stable/insights-agent/templates/right-sizer/controller-deployment.yaml
```diff
@@ -0,0 +1,35 @@
...
+          image: "{{ .Values.rightsizer.image.repository }}:{{ .Values.rightsizer.image.tag }}"
+          imagePullPolicy: {{ .Values.rightsizer.image.pullPolicy }}
+          command:
+          - "right-sizer"
+          # NOt yet supported by the controller:
+          #- "-stateconfigmap"
+          #- "{{ .Release.Namespace }}/{{ .Values.rightsizer.stateconfigmapname }}"
```
makoscafee, line 36: You will also need to add health check port ... otherwise `kubehunter` scan will be complaining

stable/insights-agent/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 description: A Helm chart to run the Fairwinds Insights agent
 name: insights-agent
 version: 1.16.4
```
makoscafee: need to update this too.

stable/insights-agent/templates/right-sizer/controller-deployment.yaml
ivanfetch-wt: > You will also need to add health check port ... otherwise `kubehunter` scan will be complaining

Making sure I understand - do you mean that I need to supply a readiness, or liveness, or both probes? The current release of the controller does not have those, but I can include one or both of them in the next release.

stable/insights-agent/templates/right-sizer/controller-deployment.yaml
ivanfetch-wt: Thanks - moved in the deployment template values specification, and defined in values also.

stable/insights-agent/templates/right-sizer/controller-deployment.yaml
makoscafee: @ivanfetch-fw there is a check to make sure both  `readiness` and `liveness` are available. including them in the next release should be fine. thanks.

rbren (approved):

PR #618, opened 2021-11-15
Add PDB to insights-admission

ivanfetch-wt:
I'd appreciate some additional eyes on the logic at the top of the PDB template -

* is this a reasonable way to enable the PDB (based on whether the HPA is enabled or replicaCount being more than 1)?

**Why This PR?**
**Changes**
Adds a PDB for the insights-admission chart. This was born from a customer that spins down a cluster, which caused there to be no admission controller pods available to facilitate workloads being spun back up.

**Checklist:**

### DIFF
--- a/stable/insights-admission/Chart.yaml
+++ b/stable/insights-admission/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: insights-admission
 description: A Validating Webhook Admission Controller that utilizes Fairwinds Insights.
 type: application
-version: 0.4.1
+version: 0.4.2
 appVersion: "0.4"
 maintainers:
 - name: rbren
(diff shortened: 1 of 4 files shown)


makoscafee:
LGTM 
you need to add `pdb.minAvailable` in  ReadMe

makoscafee (approved):

ivanfetch-wt:
I added a `pdb.enabled` chart value, so that there would be an option to opt out of the PDB.

I also clarified PDB-related chart values, and re-generated the ReadMe.

I used `helm template` to test the if statement at the top of the PDB:

* autoscaling is enabled and the minimum replicas are 1 (no PDB) or 2+ (PDB is created).
* Autoscaling is not enabled, deployment replicas are 1 (no PDB) or 2+ (PDB is created).
* Combinations of the above but `pdb.enabled` is set to false (no PDB).

makoscafee (approved):

lucasreed (approved):

PR #619, opened 2021-11-18
Disable database cleanup by default

jdesouza:
**Why This PR?**

Fixes #

**Changes**

*
*

**Checklist:**

### DIFF
--- a/stable/fairwinds-insights/ci/test-values.yaml
+++ b/stable/fairwinds-insights/ci/test-values.yaml
@@ -48,7 +48,7 @@ emailCronjob:
   schedules: []
 
 databaseCleanupCronjob:
-  enabled: false
+  enabled: true
   schedules: []
 
 minio:
--- a/stable/fairwinds-insights/ci/test-values.yaml
+++ b/stable/fairwinds-insights/ci/test-values.yaml
@@ -48,6 +48,7 @@ emailCronjob:
   schedules: []
 
 databaseCleanupCronjob:
+  enabled: true
   schedules: []
 
 minio:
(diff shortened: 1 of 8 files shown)


ivanfetch-wt (approved):
As long as our Helm values for prod are updated *first* to include `databaseCleanupCronjob.enabled=true`, this PR looks good.

I tested to make sure the template renders and uses the new value.

stable/fairwinds-insights/values.yaml
```diff
@@ -284,6 +284,8 @@ emailCronjob:
     runAsUser: 10324
 
 databaseCleanupCronjob:
+  # -- Enable database cleanup
+  enabled: false
```
makoscafee, line 288: BTW, do we need to enable this in `ci/test.yaml`?

makoscafee (approved):

stable/fairwinds-insights/values.yaml
jdesouza: that's a good idea

stable/insights-agent/ci/test-values.yaml
```diff
@@ -59,3 +59,6 @@ awscosts:
   tagkey: kubernetes_cluster
   tagvalue: staging.internal.reactiveops.com
   workgroup: cur_athena_workgroup
+
```
jdesouza: @makoscafee right?

makoscafee: yes that looks about right

stable/fairwinds-insights/ci/test-values.yaml
```diff
@@ -48,6 +48,7 @@ emailCronjob:
   schedules: []
 
 databaseCleanupCronjob:
+  enabled: true
```
makoscafee, line 51: actually just to think of it ... you do not need this. none of the cronjob are enabled here

rbren:
@jdesouza thanks for the fix! However, I think we should revert this PR, and make the change in the insights-release repo - we want to make sure our self-hosted users have this job on by default.

PR #629, opened 2021-11-30
A modified chart for the datadog-apm cluster agent 

sammc3:
A modified chart for the datadog-apm cluster agent to remediate older deployments of this agent for fairwinds.


incubator/datadog-apm/templates/cluster-agent-deployment.yaml
```diff
@@ -0,0 +1,279 @@
...
+      dnsConfig:
+{{ toYaml .Values.clusterAgent.dnsConfig | indent 8 }}
+      {{- end }}
+      containers:
+      - name: cluster-agent
+        image: "{{ .Values.clusterAgent.image.repository }}:{{ .Values.clusterAgent.image.tag }}"
+        {{- with .Values.clusterAgent.command }}
+        command: {{ range . }}
+          - {{ . | quote }}
+        {{- end }}
```
lucasreed, line 85: I think we can hard code the commands we know should run in this chart ('["trace-agent", "-config=/etc/datadog-agent/datadog.yaml"]`) and use the values provided in the `values.yml` file to add extra args if necessary.

lucasreed (changes requested):
Thanks for getting this going @sammc3! 🎉 

I have some concerns with the complexity as it looks like it's mostly copied over from the main datadog chart and there's likely a lot of things here we don't really need. For instance, are all those environment variables in the deployment required for APM functionality? Same question for the RBAC definitions.

Also we shouldn't call any of this `cluster-agent` since that's not the purpose it's serving. Any kube resources should be called some form of `datadog-apm` where you have `cluster-agent` currently.

I've made one in-line comment about the `command` to run in the containers, but if you could also pare this down to the parts that are required for APM, that would be awesome.

Lastly, has this been tested in a running cluster? We'll want to confirm that before merging this.

incubator/datadog-apm/templates/cluster-agent-deployment.yaml
```diff
@@ -182,7 +182,7 @@ spec:
                 name: {{ template "clusterAgent.tokenSecretName" . }}
                 key: token
           - name: DD_KUBE_RESOURCES_NAMESPACE
-            value: {{ .Release.Namespace }}
+            value: {{ .Values.namespace }}
```
lucasreed, line 185: We should keep this as `.Release.Namespace`

incubator/datadog-apm/templates/cluster-agent-pdb.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: policy/v1
 kind: PodDisruptionBudget
 metadata:
   name: {{ template "datadog-apm.fullname" . }}-cluster-agent
-  namespace: {{ .Release.Namespace }}
+  namespace: {{ .Values.namespace }}
```
lucasreed, line 6: We should keep this as `.Release.Namespace`

incubator/datadog-apm/values.yaml
```diff
@@ -5,7 +5,7 @@
 nameOverride: ""
 fullnameOverride: ""
 targetSystem: linux
-namespace:  # Need to override this value with datadog
+namespace: default # Need to override this value with datadog
```
lucasreed: This should be removed from the values file. We don't need to specify namespace as that's done by the helm release process.

lucasreed (changes requested):

PR #631, opened 2021-12-01
If/fwi-976  Migrate from ValidatingWebhookConfiguration v1beta1 to support k8s 1.22

ivanfetch-wt:
**Why This PR?**
Update the admission controller chart to support Kube 1.22.

**Checklist:**

### DIFF
--- a/stable/insights-admission/templates/configuration.yaml
+++ b/stable/insights-admission/templates/configuration.yaml
@@ -1,4 +1,8 @@
+{{- if .Capabilities.APIVersions.Has "admissionregistration.k8s.io/v1" }}
 apiVersion: admissionregistration.k8s.io/v1
+{{- else }}
+apiVersion: admissionregistration.k8s.io/v1beta1
+{{- end }}
 kind: ValidatingWebhookConfiguration
 metadata:
   name: {{ include "insights-admission.fullname" . }}
@@ -8,7 +12,9 @@ metadata:
     {{- end }}
 webhooks:
 - admissionReviewVersions:
+{{- if .Capabilities.APIVersions.Has "admissionregistration.k8s.io/v1" }}
   - v1
+{{- end }}
   - v1beta1
   clientConfig:
     {{- if .Values.caBundle }}
--- a/stable/insights-admission/templates/configuration.yaml
+++ b/stable/insights-admission/templates/configuration.yaml
@@ -1,4 +1,4 @@
-apiVersion: admissionregistration.k8s.io/v1beta1
+apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
   name: {{ include "insights-admission.fullname" . }}
@@ -8,7 +8,7 @@ metadata:
     {{- end }}
 webhooks:
 - admissionReviewVersions:
-  - v1beta1
+  - v1
   clientConfig:
     {{- if .Values.caBundle }}
     caBundle: {{ .Values.caBundle | quote }}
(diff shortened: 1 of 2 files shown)


stable/insights-admission/templates/configuration.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: admissionregistration.k8s.io/v1beta1
+apiVersion: admissionregistration.k8s.io/v1
```
rbren, line 2: when was v1 introduced? We might want to continue supporting the old version - I think we still have some folks as far back as 1.14

rbren: Check out `Capabilities.Has` for how to do this - we have it in a few other places in this chart

ivanfetch-wt: @rbren  While looking into this, the current master version of the chart does not work with Kube 1.14 - to me that means we don't need to support 1.14?

insights-agent install output on Kube 1.14 from chart master:
```
Release "insights-admission" does not exist. Installing it now.
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: [ValidationError(ValidatingWebhookConfiguration.webhooks[0].clientConfig.service): unknown field "port" in io.k8s.api.admissionregistration.v1beta1.ServiceReference, ValidationError(ValidatingWebhookConfiguration.webhooks[0]): unknown field "matchPolicy" in io.k8s.api.admissionregistration.v1beta1.Webhook, ValidationError(ValidatingWebhookConfiguration.webhooks[0]): unknown field "objectSelector" in io.k8s.api.admissionregistration.v1beta1.Webhook]
```

ivanfetch-wt: Update, we can support Kube 1.15 (which does currently work with our master branch). I've updated the template with if/else statements for the API version, and tested in Kube 1.15 and again in 1.22.

rbren (approved):

stable/insights-admission/templates/configuration.yaml
```diff
@@ -8,6 +12,9 @@ metadata:
     {{- end }}
 webhooks:
 - admissionReviewVersions:
+{{- if .Capabilities.APIVersions.Has "admissionregistration.k8s.io/v1" }}
+  - v1
+{{- end }}
   - v1beta1
```
rbren, line 18: should we keep this one or put it inside an `else`?

ivanfetch-wt: > should we keep this one or put it inside an    else   ?

I **believe** we want both, but am not 100%. I'm not clear on who/what may send what versions of AdmissionReview objects to our webhook. 

I looked at [this doc](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/) when creating this PR - it says this about that field:
Webhooks can specify what versions of    AdmissionReview objects they accept with the admissionReviewVersions    field in their configuration ...
Webhooks are required to support at least one AdmissionReview version understood by the current and previous API server.
API servers send the first AdmissionReview version in the admissionReviewVersions list they support. If none of the versions in the list are supported by the API server, the configuration will not be allowed to be created. If an API server encounters a webhook configuration that was previously created and does not support any of the AdmissionReview versions the API server knows how to send, attempts to call to the webhook will fail and be subject to the failure policy.

@rbren / @makoscafee  What is your interpretation of the above, and whether we want both fields?

makoscafee: I would go with leaving both fields there... I am assuming from version `1.16` to `1.21` there are both capabilities `v1beta1` and `v1`

makoscafee (approved):

stable/insights-admission/templates/configuration.yaml
ivanfetch-wt: Correct RE: 1.16-1.21.

rbren (approved):

PR #633, opened 2021-12-03
[insights-agent] change from using ebpf support for falco to falco-probe kernel module

makoscafee:
**Why This PR?**

if you look driver download page [here](https://download.falco.org/?prefix=driver/3aa7a83bf7b9e6229a3824e3fd1f4452d1e95cb4/) for Falco for our nodes kernel `5.8.0-1041-aws` you will see we only have `falco-probe` kernel module driver(with extension `.ko`) but no `eBPF` probe (with extension `.o`) which means enabling ebpf module will sometimes not work. Reverted to default Falco installation behaviour.

Fixes #

**Changes**

*
*

**Checklist:**

### DIFF
--- a/stable/insights-agent/Chart.yaml
+++ b/stable/insights-agent/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to run the Fairwinds Insights agent
 name: insights-agent
-version: 1.16.4
+version: 1.16.5
 appVersion: 6.5.1
 maintainers:
   - name: rbren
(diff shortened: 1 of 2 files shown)


ivanfetch-wt (approved):
@makoscafee It sounds like this change does not enable EBPF by default - would that be a significant enough change RE: the chart version to also bump the minor number?

sammc3 (approved):
LGTM.

makoscafee:
> @makoscafee It sounds like this change does not enable EBPF by default - would that be a significant enough change RE: the chart version to also bump the minor number?

It does make sense ... but since nobody is using Falco at the moment and this change is not going to break anything, a major version bump should be fine i think.

rbren (approved):

PR #634, opened 2021-12-03
[insights-agent] Overide fullname to be able to work with sidekick

makoscafee:
**Why This PR?**

Fixes #

**Changes**

*
*

**Checklist:**


stable/insights-agent/templates/falco/rbac.yaml
```diff
@@ -13,10 +13,31 @@ metadata:
...
+      - 'get'
+      - 'list'
+  - apiGroups:
+      - 'batch'
+    resources:
+      - 'jobs'
+      - 'cronjobs'
+    verbs:
+      - 'get'
+      - 'list'
```
rbren: I think we should make this the `view` clusterrole

makoscafee: mmh will `view` able to solve this access issue ? `cannot list resource \"jobs\" in API group \"batch\" in the namespace \"xxx\""`

rbren (approved):

PR #642, opened 2021-12-09
[fairwinds-metrics] update values, README, and use helm-docs

lucasreed:
**Why This PR?**
To update the values needed for fairwinds-metrics and support silencing
the istio metric

**Changes**

- removed all listener parts as that functinality does not exist in
  custom-metrics any longer.
- add suppoert to add environment vars and specifically turn off the
  istio monitor
  - added docs for this
- start using helm-docs to generate README

Fixes
- removed all the stuff for listener

**Checklist:**


incubator/fairwinds-metrics/README.md
```diff
@@ -1,45 +1,42 @@
...
-| `listener.ingress.tls` | listener ingress tls config | | |
-| `listener.secret.create` | When set to true, creates the listener secret. | `true` | |
-| `listener.secret.name` | When `listener.secret.create` is set to `false`, indicates the secret to mount as environment variables to the container. | `` | |
-| `listener.service.type` | | `ClusterIP` | | |
-| `listener.service.port` | | | `8080` | |
+## Disabling certain metrics
+
+As of chart version 0.6.0 you can disable metrics by using environment variables.
+
+If you set any of the following pod env variables to `"false"` then the metric will be disabled. By default they are all enabled:
```
sudermanjr, line 12: Based on this paragraph and the values, I'm unclear on how to do this. 

I think we set `controller.env` to something specific? I think we could use a little more clarity here

sudermanjr (changes requested):

incubator/fairwinds-metrics/README.md
lucasreed: Yeah that wasn't super clear. I put an example in my recent commit that I think shows it a little better.

sudermanjr (approved):

PR #665, opened 2022-01-11
[insights-agent] update admission control

rbren:
**Why This PR?**
Update the admission controller sub-chart

**Changes**

* Unpin the patch version of Fairwinds-managed charts
* Remove the requirements.lock in order to always update the fairwinds-managed patch versions


stable/insights-agent/requirements.lock
```diff
@@ -1,15 +0,0 @@
-dependencies:
```
makoscafee, line 1: mmh why was this file deleted?

makoscafee (approved):

stable/insights-agent/requirements.lock
rbren: We unpinned the versions so that we don't have to keep updating this chart every time a downstream has a patch.

sudermanjr (approved):

PR #677, opened 2022-01-19
[Polaris] Update replicas for polaris dashboard and webhook deployment

makoscafee:
**Why This PR?**
This PR update dashboard and webhook deployment replicas to 2 to match new added rule enabled by default in polaris `deploymentMissingReplicas`

Fixes #
- Fixes `deploymentMissingReplicas` check error 
**Changes**

* Update replicas for webhook and dashboard to `2`
*

**Checklist:**


stable/polaris/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 description: Validation of best practices in your Kubernetes clusters
 name: polaris
-version: 4.2.3
-appVersion: "4.2"
+version: 5.0.0
+appVersion: "5.0"
```
sudermanjr, line 5: I don't think the version of polaris is changing, so we should keep this the same I think

stable/polaris/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 description: Validation of best practices in your Kubernetes clusters
 name: polaris
-version: 4.2.3
-appVersion: "4.2"
+version: 5.0.0
```
sudermanjr, line 4: This doesn't seem like a major change to the chart. Honestly it feels more like a patch version to me

sudermanjr (changes requested):

rbren (approved):

rbren:
Should we try updating kube-hunter to get rid of that vuln?

makoscafee:
> Should we try updating kube-hunter to get rid of that vuln?

Sure, I can do that in this PR.

PR #684, opened 2022-01-25
Add fleet install test

mhoss019:
**Why This PR?**
Adding fleet install test for insights-agent

Fixes #

**Changes**

* New Fleet install test run if the `insights-agent` chart changes

**Checklist:**

* [ ] Any new values have been added to the README for the Chart, or `helm-docs --sort-values-order=file` has been run for the charts that support it.

### DIFF
--- a/scripts/fleet-install-test.sh
+++ b/scripts/fleet-install-test.sh
@@ -1,0 +1,45 @@
+#!/bin/sh
+
+set -o errexit
+set -o nounset
+set -x
+
+# Check whether insights-agent is part of the changed charts
+cd /charts
+CHANGED="$(ct list-changed --config ./scripts/ct.yaml)"
+
+case "$CHANGED" in 
+  *insights-agent*)
+    printf "The changed charts include insights-agent. Running fleet install test\n"
+
+    # We use the be-main server to do the test. This is not ideal but will suffice for now
+    # Here we check to make sure that the server is up
+    URL="https://be-main.k8s.insights.fairwinds.com" 
+    retry=0
+    while ! curl -k --no-progress-meter $URL; do
+      printf "Server is not up yet. Waiting for it to start...\n";
+      sleep 20
+
+      retry=$(( retry + 1 ))
+      if [ $retry -gt 30 ]; then
+        printf "Unable to curl the server. Giving up on it.\n"
+        exit 1
+      fi
+    done
+
+    # Install insights-agent using the fleet install method
+    random_number=$(shuf -i 0-500 -n1)
+    cluster_name="test-fleet-$random_number"
+    printf "\nServer is up. Running Fleet install for cluster %s\n" "$cluster_name"
+    
+    helm dependency build ./stable/insights-agent
+    helm upgrade --install insights-agent ./stable/insights-agent -f ./stable/insights-agent/ci/fleet-install-test.yaml \
+      --namespace insights-agent \
+      --create-namespace \
+      --set insights.cluster="$cluster_name" 
+
+    kubectl wait --for=condition=complete job/fleet-installer --timeout=120s --namespace insights-agent
+    kubectl wait --for=condition=complete job/polaris --timeout=120s --namespace insights-agent
+    kubectl wait --for=condition=complete job/workloads --timeout=120s --namespace insights-agent
+    ;;
+esac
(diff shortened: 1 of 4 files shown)


scripts/fleet-install-test.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/sh
+
+# Check whether insights-agent is part of the changed charts
+cd /charts
+CHANGED="$(ct list-changed --config ./scripts/ct.yaml)"
```
mhoss019, line 9: We only run the fleet install test when there is a change in `insights-agent` chart, is that reasonable?

scripts/fleet-install-test.sh
```diff
@@ -0,0 +1,47 @@
...
+    helm dependency build ./stable/insights-agent
+    helm upgrade --install insights-agent ./stable/insights-agent \
+      --namespace insights-agent \
+      --create-namespace \
+      --set insights.organization="acme-co" \
+      --set insights.cluster=$cluster_name \
+      --set insights.host=$URL \
+      --set fleetInstall=true  \
+      --set insights.tokenSecretName=insights-token \
+      --set insights.apiToken="thisisanadmintoken" \
```
mhoss019: Have the token listed here. Is that okay?

scripts/fleet-install-test.sh
```diff
@@ -0,0 +1,47 @@
...
+      --set polaris.enabled=true
+
+kubectl wait --for=condition=complete job/fleet-installer --timeout=120s --namespace insights-agent
+kubectl wait --for=condition=complete job/polaris --timeout=120s --namespace insights-agent
```
mhoss019: We only have polaris and workloads running since just installing the Fleet Install specifically. Does that make sense?

stable/insights-agent/values.yaml
```diff
@@ -16,7 +16,7 @@ uploader:
   resources:
     limits:
       cpu: 250m
-      memory: 512Mi
+      memory: 128Mi
```
mhoss019: Will put this back when I get approval. This was for testing

mhoss019:
Currently this will run no matter what. However, if the `insights-agent` chart does not change, it will just quickly exit the script. However, the initial part of the script where it sets up the kind cluster + the command_runner container still happens.
I tried to make it so that we have a check beforehand such that "If the insights-agent chart changed, THEN do the testing" but circleci doesn't really have that kind of logic.
Any thoughts on how to do this better?

scripts/fleet-install-test.sh
rbren: Yes I think having the debug token here is OK

scripts/fleet-install-test.sh
rbren: Yes agree

scripts/fleet-install-test.sh
rbren: I expected these values to be set in `./stable/insights-agent/ci/fleet-install-values.yaml`. It'll probably be easier to maintain there

scripts/fleet-install-test.sh
rbren: Yes definitely desireable

.circleci/config.yml
```diff
@@ -134,6 +134,12 @@ workflows:
           executor:
             name: ci-images-large
           <<: *kind_configuration_helm3
+      - rok8s11/kubernetes_e2e_tests:
```
rbren: Instead of starting a whole new CircleCI job, I think we should leverage the existing test script. That way the fleet install runs on each version of Kubernetes we support, not just 1.21

IIRC you mentioned that just adding another YAML file to `stable/insights-agent/ci` caused `ct` to run a separate test with that file - is that true?

If not maybe we can modify [this script](https://github.com/FairwindsOps/charts/blob/master/scripts/e2e-test.sh) to support it

mhoss019: @rbren that's correct. If I add a *-values.yaml to the `stable/insights-agent-ci` folder, it will run the regular ct tests. Basically it will install a previous version of the chart with those *values.yaml and upgrade it to the new version. Then just install the new version.

My concerns are:
* The cluster name will in that case be static
* We can't do a check to make sure the be-main server is ready
* We can't do the kubectl wait for completion checks -> Not until later when I figure out how to do this

Is that what we want? 
I can still add this script to be run as part of the `e2e` tests so it runs for every kubernetes cluster. But it won't be through charts-testing.

rbren: 👍 makes sense! I like the solution you have here.

rbren (approved):

PR #687, opened 2022-01-27
A modified datadog-apm cluster agent chart

sammc3:
A modified datadog-apm cluster agent chart to remediate older deployments of this agent for Fairwinds ClusterOps.


incubator/datadog-apm/values.yaml
```diff
@@ -0,0 +1,80 @@
...
+      cpu: 50m
+      memory: 150Mi
+    requests:
+      cpu: 50m
+      memory: 150Mi
+  hpa:
+    enabled: true
+    minReplicas: 1
+    maxReplicas: 6
+    averageCpuUtilization: 30
```
lucasreed: I think we can default to a higher cpu utilization and also have `enabled` default to false. Since we'll likely be the main users of this chart, I don't think we need an HPA for this.

lucasreed (changes requested):
Looks good but a comment on a couple of the HPA defaults. @sudermanjr thoughts on that?

incubator/datadog-apm/values.yaml
sudermanjr: I definitely think we should default to using the HPA. Part of the reason for running this chart is the scaling functionality for us. 

However, I think we need a memory-based autoscaler, since that's our usual failure point

sammc3: To confirm a memory based hpa vs a cpu based hpa or should we have both? Also what avg memory utilization should we target?

incubator/datadog-apm/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+# Chart is a subset of the datadog chart v2.4.25
+apiVersion: v2
+name: datadog-apm
+description: A modified chart that only installs the datadog-apm agent
+type: application
+version: 0.0.1
+appVersion: "1.0.0"
```
lucasreed: Can we make this equal to the datadog application? The image tag specified is `7.23.0`

lucasreed (changes requested):
One small change, then I think it's good to go

incubator/datadog-apm/Chart.yaml
sammc3: Done.

lucasreed (approved):

incubator/datadog-apm/values.yaml
sammc3: Modified HPA cpu to memory based.

PR #691, opened 2022-02-02
refactor download script

rbren:
**Why This PR?**
This adds some debug info to the download script, and deduplicates logic across two jobs.

**Changes**
* Dedupe script
* Fix test-pod
* Add debug info

**Checklist:**

### DIFF
--- a/stable/insights-agent/download-kubectl.sh
+++ b/stable/insights-agent/download-kubectl.sh
@@ -1,0 +1,29 @@
+#! /bin/sh
+
+set -e
+
+mkdir /tmp/bin
+export PATH=$PATH:/tmp/bin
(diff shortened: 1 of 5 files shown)


stable/insights-agent/download-kubectl.sh
```diff
@@ -0,0 +1,29 @@
+#! /bin/sh
+
+set -e
```
mhoss019, line 3: I think this needs to be `set -eo pipefail`
Otherwise the pipeline will still not fail

ivanfetch-wt (approved):
Ref: https://fairwindsops.slack.com/archives/CNQBGRGFN/p1643824334147639

PR #694, opened 2022-02-03
[insights-agent] Fix version parsing

rbren:
**Why This PR?**
Adds some debug info to figure out an issue in 1.22, and tries to use a better field

**Checklist:**

### DIFF
--- a/stable/insights-agent/download-kubectl.sh
+++ b/stable/insights-agent/download-kubectl.sh
@@ -16,7 +16,7 @@ curl -fsSLo jq https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux6
 
 echo "Getting the Kubernetes version from the API. . ."
 kube_version_info=$(curl -kfsSL https://kubernetes.default.svc/version?timeout=32s)
-echo "Found Kubernetes version info: $kube_version"
+echo "Found Kubernetes version info: $kube_version_info"
 git_version=$(echo "${kube_version_info}" | jq -r .gitVersion | cut -d- -f1)
 echo "Git version is ${git_version}"
 kube_minor_version=$(echo "${kube_version_info}" | jq -r .minor)
--- a/stable/insights-agent/download-kubectl.sh
+++ b/stable/insights-agent/download-kubectl.sh
@@ -25,7 +25,7 @@ echo "Minor version is ${kube_minor_version}"
 
 version_pattern='v[0-9]+\.[0-9]+.[0-9]+)'
 kubectl_version=""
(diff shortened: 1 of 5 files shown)


stable/insights-agent/download-kubectl.sh
```diff
@@ -1,29 +1,41 @@
 #! /bin/sh
 
-set -e
+set -eo pipefail
```
mhoss019, line 3: I tried adding the `-o pipeline` here but it doesn't seem to like it. It doesn't error out but it says "Invalid option" or something

rbren: @mhoss019 it's `pipefail` not `pipeline` 😄 

Seems to be OK now - anything I should check besides that the tests are passing?

mhoss019: Yeah it doesn't like this. If you look at the logs it has "illegal option"

mhoss019: OOOOOO my bad! Looks good. Thanks!

rbren: That's from the previous run 😆 

```
sh: set: line 1: illegal option -o pipeline
```
now it's `pipefail`

stable/insights-agent/download-kubectl.sh
```diff
@@ -1,29 +1,41 @@
...
 echo "Downloading jq . . ."
-curl -sSLo jq https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 && chmod +x jq
+curl -fsSLo jq https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 && chmod +x jq
+
 echo "Getting the Kubernetes version from the API. . ."
-kube_version=$(curl -ksS https://kubernetes.default.svc/version?timeout=32s | jq -r .gitVersion | cut -d- -f1)
-echo "Found Kubernetes version $kube_version"
-kube_minor_version=$(echo $kube_version |cut -d. -f2)
+kube_version_info=$(curl -kfsSL https://kubernetes.default.svc/version?timeout=32s)
+echo "Found Kubernetes version info: $kube_version"
```
mhoss019: where is $kube_version defined here?

mhoss019: I think you want $kube_version_info

rbren: good catch!

mhoss019 (approved):

makoscafee (approved):

PR #707, opened 2022-02-14
Update admission plugin version  in admission chart

makoscafee:
**Why This PR?**

Fixes #

**Changes**

* Update appVersion
*

**Checklist:**

### DIFF
--- a/stable/insights-admission/Chart.yaml
+++ b/stable/insights-admission/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: insights-admission
 description: A Validating Webhook Admission Controller that utilizes Fairwinds Insights.
 type: application
-version: 1.0.0
-appVersion: "0.5"
+version: 1.0.1
+appVersion: "1.1"
 maintainers:
 - name: rbren
 - name: makoscafee


ivanfetch-wt:
@makoscafee  I have a few questions about this rollout:

* When do we expect users to start using this new chart version? I'm thinking about users that pull later charts before we have specified a newer version in our install-hub UI.
* Are we also ready to include PR #596 by bumping to this version of admission? That PR adds an `insights info` function to rego policies, and updates some error handling / returning.

makoscafee:
> @makoscafee I have a few questions about this rollout:
> 
> * When do we expect users to start using this new chart version? I'm thinking about users that pull later charts before we have specified a newer version in our install-hub UI.
> * Are we also ready to include PR [[insights-agent] update admission-controller #596](https://github.com/FairwindsOps/charts/pull/596) by bumping to this version of admission? That PR adds an `insights info` function to rego policies, and updates some error handling / returning.

* Ideally we are expecting people to use this once we ship agent `2.0` unless use is using `insights-admission` separately and upgrade to this version which is version 1.0.1. most people use it through the agent and that should be fine.
* After this PR is merged will update the agent chart with this new admission chart dependency and the chart will be merged only when we are about to release new release.

rbren (approved):

PR #752, opened 2022-03-17
[stable/vpa] Allow specifying custom key/paths for the vpa-tls-cert

### LINKED ISSUE #751: [VPA] Allow specifying custom key/paths for the vpa-tls-cert
This applies to the Vertical-pod-autoscaler chart.

I have an environment that doesn't have internet access, which means the admission-controller's **certgen** job is unable to do certain operations like [here](https://github.com/FairwindsOps/charts/blob/master/stable/vpa/templates/admission-controller-certgen.yaml#L81) and thus is unable to create the `vpa-tls-certs` secret. Instead I want to use [cert-manager](https://cert-manager.io/) to generate the certificates secret for the VPA.

However; there is a mismatch between the paths where the VPA expects the certs on its file-system, and the secret generated by cert-manager. Since the secret itself is just mounted straight up with no [configuration available](https://github.com/FairwindsOps/charts/blob/d83ff3fe24415ed810f4654215a0204f8e662b06/stable/vpa/templates/admission-controller-deployment.yaml#L83), it means the secret must follow a certain format so that all of the certs and keys get mounted in the right paths.

Given how the VPA mounts the `vpa-tls-certs` secret, the secret is forced to look like:

```yml
apiVersion: v1
kind: Secret
metadata:
  name: vpa-tls-certs
data:
  caCert.pem: <>
  caKey.pem: <>
  serverCert.pem: <>
  serverKey.pem: <>
```

Meanwhile, cert-managers [Certificate type](https://cert-manager.io/docs/concepts/certificate/) generates `kubernetes/tls` type secrets, which means they always look like:

```yml
apiVersion: v1
kind: Secret
metadata:
  name: vpa-tls-certs
data:
  ca.crt: <>
  tls.crt: <>
  tls.key: <>
```

Given the answer to [this issue](https://github.com/cert-manager/cert-manager/issues/675), cert-manager doesn't want to allow people to specify custom keys which forces the problem on downstream users. Because of this mismatch, there is no way at present to use the VPA chart with cert-manager generated certificates.

I would like a field in the values file that allows specifying the `items` key/paths on the `vpa-tls-certs`, like in [this comment](https://github.com/cert-manager/cert-manager/issues/675#issuecomment-713621568).

Such a field would allow a user to map an arbitrary secret to the paths expected by the admission-controller pod. This solve by above issue like so:

```yml
admissionController:
  tlsSecretKeys:
    - key: ca.crt
      path: caCert.pem
    - key: tls.crt
      path: serverCert.pem
    - key: tls.key
      path: serverKey.pem
    # CaKey isn't necessary for the admission-controller to function
```

An alternative is to modify the **Vertical pod autoscaler** itself to accept arguments for where to find each of the certificates it looks for, instead of always being `/etc/caCert.pem` etc. Then we would modify this helm chart to have an `extraArgs`, a common pattern seen in many charts. This would make configuring the admission-controller with an arbitrary secret *possible*, but still kind of cumbersome.

In terms of viability I feel any solution that doesn't involve modifying upstream Kubernetes is preferable, and probably a lot easier to get merged in given the strict contribution guidelines. I also think the other solution is cleaner

**Additional context**

I have an PR I will submit that adds customizable key paths. Thanks for the awesome chart.

aperullo:
Closes #751 

**Why This PR?**
See #751 for the full details.

This PR allows certificates created by means other than the admission-controllers [certgen](https://github.com/FairwindsOps/charts/blob/master/stable/vpa/templates/admission-controller-certgen.yaml) job to be mapped into the admission-controller pod. As it stands, any secret not following the format created by certgen will not be usable. This includes the official [kubernetes/tls-secret](https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets) format. 

**Changes**

* Add a `tlsSecretKeys` field to allow mapping secret keys to their correct paths within the container. 
* Include the example commented out, a common practice in helm charts, that shows what the correct paths are. 
* Default this new field to empty to maintain backwards-compatibility.

**Checklist:**

### DIFF
--- a/stable/vpa/Chart.yaml
+++ b/stable/vpa/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: vpa
 description: A Helm chart for Kubernetes Vertical Pod Autoscaler
 type: application
-version: 1.3.2
+version: 1.3.0
 appVersion: 0.10.0
 maintainers:
   - name: sudermanjr
--- a/stable/vpa/Chart.yaml
+++ b/stable/vpa/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: vpa
 description: A Helm chart for Kubernetes Vertical Pod Autoscaler
 type: application
-version: 1.2.2
+version: 1.3.2
 appVersion: 0.10.0
 maintainers:
   - name: sudermanjr
--- a/stable/vpa/README.md
+++ b/stable/vpa/README.md
@@ -124,6 +124,7 @@ recommender:
 | admissionController.podSecurityContext.runAsUser | int | `65534` |  |
 | admissionController.securityContext | object | `{}` | The security context for the containers inside the admission controller pod |
 | admissionController.resources | object | `{"limits":{"cpu":"200m","memory":"500Mi"},"requests":{"cpu":"50m","memory":"200Mi"}}` | The resources block for the admission controller pod |
+| admissionController.tlsSecretKeys | list | `[]` | The keys in the vpa-tls-certs secret to map in to the admission controller |
 | admissionController.nodeSelector | object | `{}` |  |
 | admissionController.tolerations | list | `[]` |  |
 | admissionController.affinity | object | `{}` |  |
--- a/stable/vpa/templates/admission-controller-deployment.yaml
+++ b/stable/vpa/templates/admission-controller-deployment.yaml
@@ -81,6 +81,10 @@ spec:
         - name: tls-certs
           secret:
             secretName: vpa-tls-certs
+            {{- with .Values.admissionController.tlsSecretKeys }}
+            items:
+              {{- toYaml . | nindent 14 }}
+            {{- end }}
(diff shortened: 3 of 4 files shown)


stable/vpa/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: vpa
 description: A Helm chart for Kubernetes Vertical Pod Autoscaler
 type: application
-version: 1.2.2
+version: 1.3.2
```
sudermanjr:
```suggestion
version: 1.3.0
```

sudermanjr (changes requested):
Thanks for the contribution!

I left a small nit on the version number, but other than that seems great.

stable/vpa/Chart.yaml
aperullo: Whoops that was silly of me. Fixed

sudermanjr (approved):

PR #781, opened 2022-04-22
[stable/goldilocks] fixing the KubeVersion capability with EKS clusters

Alizandieh:
**Why This PR?**
I had an issue with ingress resource on EKS clusters. the .Capabilities.KubeVersion.Version in EKS cluster returns something like 1.21.1-eks-ghtsn 

Fixes #

**Changes**

replacing the dot with a dash in the version number could get the semverCompare function works properly. 
like having v1.19-0 instead of v1.19.0
It's been tested and works fine. 

**Checklist:**


lucasreed (commented):
Will it work if you simply remove the patch version completely and check for anything greater than v1.19? Maybe a better question is: do we even need to check semver of the cluster version here? If the v1 API version is available, that seems like the only check we should need. @sudermanjr any idea if we need to check the kube version here?

Alizandieh:
> Will it work if you simply remove the patch version completely and check for anything greater than v1.19? Maybe a better question is: do we even need to check semver of the cluster version here? If the v1 API version is available, that seems like the only check we should need. @sudermanjr any idea if we need to check the kube version here?
--------------------------------------------

That's a good point. probably checking K8s semVersion is for an extra safety. they're doing the same in Prometheus stack chart here: https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/_helpers.tpl#L138

sudermanjr (approved):

lucasreed (approved):
We chatted and think this will be fine to merge as-is, thanks again!

PR #792, opened 2022-04-28
[fairwinds-insights] Fix cronjobs

jdesouza:
**Why This PR?**

Fixes #

**Changes**

*
*

**Checklist:**

### DIFF
--- a/stable/fairwinds-insights/templates/_helpers.tpl
+++ b/stable/fairwinds-insights/templates/_helpers.tpl
@@ -107,11 +107,3 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{- define "secrets_credentials" -}}
 "fwinsights-timescale"
 {{- end -}}
-
-{{- define "fairwinds-insights-private.cronjobImageTag" -}}
-{{- if  .Values.fairwindsInsights.cronjobImage.tag  -}}
-{{- .Values.fairwindsInsights.cronjobImage.tag -}}
-{{- else -}}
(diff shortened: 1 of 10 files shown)


stable/fairwinds-insights/values.yaml
```diff
@@ -72,6 +72,7 @@ options:
     cpu: 1000m
 
 hubspotCronjob:
+  enabled: false
```
rbren: rather than creating a new `enabled` field, we can just set `schedules: []`

stable/fairwinds-insights/templates/_helpers.tpl
```diff
@@ -107,3 +107,11 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{- define "secrets_credentials" -}}
 "fwinsights-timescale"
 {{- end -}}
+
+{{- define "fairwinds-insights-private.cronjobImageTag" -}}
```
rbren: I don't think we want to call this `fairwinds-insights-private`

rbren: We should be able to just reuse `cronjobImageTag` from above

jdesouza: done

stable/fairwinds-insights/values.yaml
jdesouza: done

rbren (approved):
Thanks James!

PR #799, opened 2022-04-29
 [stable/polaris] Added mutating webhook configuration

makoscafee:
**Why This PR?**
This adds mutating webhook configurations
Fixes #

**Changes**

*
*

**Checklist:**


stable/polaris/values.yaml
```diff
@@ -104,6 +104,8 @@ webhook:
   objectSelector: {}
   # webhook.rules -- An array of additional for the ValidatingWebhookConfiguration. Each requires a set of apiGroups, apiVersions, operations, resources, and a scope.
   rules: []
+  # webhook.mutatingRules -- An array of additional for the MutatingWebhookConfiguration. Each requires a set of apiGroups, apiVersions, operations, resources, and a scope.
```
ivanfetch-wt: I think this is missing the word rules, it states `of additional for`. This also looks to be the case for the validating web hook value, which may have been copied for this new value.

stable/polaris/templates/mutate-webhook.configuration.yaml
```diff
@@ -0,0 +1,32 @@
...
+  clientConfig:
+    {{- if .Values.webhook.caBundle }}
+    caBundle: {{ .Values.webhook.caBundle | quote }}
+    {{- end }}
+    service:
+      name: {{ include "polaris.fullname" . }}-webhook
+      namespace: {{ .Release.Namespace }}
+      path: /mutate
+      port: 443
+  name: polaris.fairwinds.com
```
ivanfetch-wt, line 25: The [mutating Webhooks spec](https://kubernetes.io/docs/reference/kubernetes-api/extend-resources/mutating-webhook-configuration-v1/) looks like it also supports the `failurePolicy` and `matchPolicy` present in our ValidatingWebhookConfiguration. Should we also support these for the mutating Webhook (using either re-using the validating Webhook values, or adding new chart values for the mutating Webhook)?

ivanfetch-wt (approved):
I commented about `failurePolicy` and `matchPOlicy` we may want to add to mutating Webhooks values, and a typo. These don't block us releasing this though.

ivanfetch-wt (approved):

rbren (approved):

PR #808, opened 2022-05-06
add value to specify apiversion for rbac settings and set value to v1

transient1:
**Why This PR?**
Kubernetes `1.22` removes `apiVersion` `rbac.authorization.k8s.io/v1beta1 `. This is a change for [incubator/oom-event-generator]

Fixes #

**Changes**

*set to v1 in rbac rolebinding and clusterrole
*

**Checklist:**


sudermanjr (changes requested):
I think we should just set this to `v1`. It's been available since 1.17 I think

incubator/oom-event-generator/templates/clusterrole.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/{{ .Values.rbac.apiVersion }}
```
rbren: I would suggest something like this instead (there are a few other examples of this around the charts repo)
```suggestion
{{- if .Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1" }}
apiVersion: rbac.authorization.k8s.io/v1
{{- else }}
apiVersion: rbac.authorization.k8s.io/v1beta1
{{- end }}
```

transient1: I hard-coded it because Andy got to me first!

sudermanjr (approved):

rbren:
Makes sense to me! We should probably change for all the incubator charts - there's a few more it looks like

rbren:
Dunno why helm-docs is angry, but if you run `helm-docs --sort-values-order=file` it should fix it

PR #818, opened 2022-05-19
[stable/insights-agent] Fix Falco proxy indention and add proxy option for admission controller

### LINKED ISSUE #817: [insights-agent]: Problems with Proxy Values
### What happened?

There are some problems when setting the proxy settings via insights-agent. 

## Falco Proxy Environment does not render valid yaml

Does not result in valid YAML:

https://github.com/FairwindsOps/charts/blob/master/stable/insights-agent/templates/falco/deployment.yaml#L62

Suggested fix:

```
env: {{- include "proxy-env-spec" . | nindent 10  }}
```

## Admission Controller missed Proxy functionality

Since the admission controller is missing the proxy env entirely

* https://github.com/FairwindsOps/charts/blob/master/stable/insights-admission/templates/deployment.yaml#L64

So therefor its kinda stuck. 

i guess you are still working on full proxy support but here my to cents:

  * You should but the settings under the `$.Values.global` key. With this you can much easier inherit the settings over subcharts. [Read more about globals](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/)

### What did you expect to happen?

Working Proxy Settings. Would make the onboard process for customers much easier.

### How can we reproduce this?

Setting Proxy Settings via 

### Version

2.0.*

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Additional context

For anyone else having troubles with kyverno in place, there's a handy kyverno policy as shortcut:

```
apiVersion: kyverno.io/v1
kind: policy
metadata:
  name: fair-inject-proxy
spec:
  background: false
  failurePolicy: Fail
  rules:
  - match:
      any:
      - resources:
          kinds:
          - Pod
    mutate:
      patchStrategicMerge:
        spec:
          initContainers:
          - (name): '*'
            env:
            - name: HTTP_PROXY
              value: "proxy"
            - name: HTTPS_PROXY
              value: "proxy"
            - name: NO_PROXY
              value: "no_proxy"
          containers:
          - (name): '*'
            env:
            - name: HTTP_PROXY
              value: "proxy"
            - name: HTTPS_PROXY
              value: "proxy"
            - name: NO_PROXY
              value: "no_proxy"
    name: inject-proxy
  validationFailureAction: audit
```

Create the namespace manually, apply this policy, then deploy agent helm-chart

makoscafee:
**Why This PR?**
Fixes Falco proxy indention issue and adds a proxy option for admission controller

Fixes #817 

**Changes**

* Change falco proxy indention from 12 to 8 to have valid yaml
* Added changed the proxy setting to global values so sub-charts can use them.
* Added proxy option to the admission controller.

**Checklist:**

### DIFF
--- a/stable/insights-agent/templates/test/test-deployment.yaml
+++ b/stable/insights-agent/templates/test/test-deployment.yaml
@@ -24,7 +24,7 @@ spec:
           env:
           {{ include "proxy-env-spec" . | indent 10 | trim }}
           securityContext:
-            {{- toYaml .Values.securityContext | nindent 12 }}
+            {{- toYaml .Values.test.securityContext | nindent 12 }}
           image: "{{ .Values.test.image.repository }}:{{ .Values.test.image.tag }}"
           imagePullPolicy: {{ .Values.test.image.pullPolicy }}
           ports:
--- a/stable/insights-admission/Chart.yaml
+++ b/stable/insights-admission/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: insights-admission
 description: A Validating Webhook Admission Controller that utilizes Fairwinds Insights.
 type: application
-version: 1.1.1
+version: 1.1.2
 appVersion: "1.1"
 icon: https://raw.githubusercontent.com/FairwindsOps/charts/master/stable/insights-admission/icon.png
 maintainers:
--- a/stable/insights-admission/README.md
+++ b/stable/insights-admission/README.md
@@ -37,6 +37,10 @@ rules:
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
+| global.proxy.http | string | `nil` | Annotations used to access the proxy servers(http). |
+| global.proxy.https | string | `nil` | Annotations used to access the proxy servers(https). |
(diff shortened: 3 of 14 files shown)


stable/insights-agent/templates/test/test-deployment.yaml
```diff
@@ -24,7 +24,7 @@ spec:
           env:
           {{ include "proxy-env-spec" . | indent 10 | trim }}
           securityContext:
-            {{- toYaml .Values.securityContext | nindent 12 }}
+            {{- toYaml .Values.test.securityContext | nindent 12 }}
```
ivanfetch-wt, line 27: Do we want this new value included in values.yaml for documentation? I see it does have other `test.*` values there.

ivanfetch-wt (approved):
Only one non-blocking question about also including the test.securityContext in the main values.yaml file.

stable/insights-agent/templates/test/test-deployment.yaml
makoscafee: Good question... I can add that in values.yaml

PR #828, opened 2022-05-25
Adding BasePath functionality to [stable/goldilocks]

lukerogers:
**Why This PR?**
If my other PR to Goldilocks (https://github.com/FairwindsOps/goldilocks/pull/471) is approved, this would extend the functionality of adding a BasePath to Goldilocks. I didn't update the chart version as the code hasn't been merged yet and needs a new release version before updating.

**Changes**
Changes proposed in this pull request:

*Add BasePath functionality

**Checklist:**

* [ X ] I have included the name of the chart in the title of this PR in square brackets i.e. `[stable/goldilocks]`.
* [ ] I have updated the chart version in `Chart.yaml` following Semantic Versioning.
* [ X ] Any new values are backwards compatible and/or have sensible default.
* [ X ] Any new values have been added to the README for the Chart, or `helm-docs --sort-values-order=file` has been run for the charts that support it.

PR #878, opened 2022-07-16
[stable/goldilocks] Version Bumps, probe fix when base path set

lukerogers:
**Why This PR?**
1. The base path functionality for [#185](https://github.com/FairwindsOps/goldilocks/issues/185) hasn't been released yet, this assumed it will go out in 4.3.4, so this shouldn't be merged until then
2. The path to the probe was incorrect (my bad), this fixes previous PR #828 

**Changes**
*Add logic to ensure the base path has correct formatting for health probe

**Checklist:**

* [ X ] I have included the name of the chart in the title of this PR in square brackets i.e. `[stable/goldilocks]`.
* [ X ] I have updated the chart version in `Chart.yaml` following Semantic Versioning.
* [ X ] Any new values are backwards compatible and/or have sensible default.
* [ X ] Any new values have been added to the README for the Chart, or `helm-docs --sort-values-order=file` has been run for the charts that support it.

### DIFF
--- a/stable/goldilocks/templates/dashboard-deployment.yaml
+++ b/stable/goldilocks/templates/dashboard-deployment.yaml
@@ -74,11 +74,11 @@ spec:
               protocol: TCP
           livenessProbe:
             httpGet:
-              path: {{ include "goldilocks.basePath" . }}/health
+              path: {{ trimSuffix "/" .Values.dashboard.basePath }}/health
               port: http
           readinessProbe:
             httpGet:
-              path: {{ include "goldilocks.basePath" . }}/health
+              path: {{ trimSuffix "/" .Values.dashboard.basePath }}/health
               port: http
           resources:
             {{- toYaml .Values.dashboard.resources | nindent 12 }}
--- a/stable/goldilocks/templates/dashboard-deployment.yaml
+++ b/stable/goldilocks/templates/dashboard-deployment.yaml
@@ -74,11 +74,11 @@ spec:
               protocol: TCP
           livenessProbe:
             httpGet:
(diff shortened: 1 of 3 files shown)


stable/goldilocks/templates/dashboard-deployment.yaml
```diff
@@ -74,11 +74,11 @@ spec:
               protocol: TCP
           livenessProbe:
             httpGet:
-              path: {{.Values.dashboard.basePath | default "/" }}health
+              path: {{ include "goldilocks.basePath" . }}/health
```
rbren:
```suggestion
              path: {{ trimSuffix "/" .Values.dashboard.basePath }}/health
```

I think this is simpler (no need for the template). It's also closer to the current logic (which expects a `/` prefix for this value)

stable/goldilocks/templates/dashboard-deployment.yaml
```diff
@@ -74,11 +74,11 @@ spec:
...
               port: http
           readinessProbe:
             httpGet:
-              path: {{.Values.dashboard.basePath | default "/" }}health
+              path: {{ include "goldilocks.basePath" . }}/health
```
rbren:
```suggestion
              path: {{ trimSuffix "/" .Values.dashboard.basePath }}/health
```

lukerogers: That is way better, thanks for the suggestion!

sudermanjr:
My apologies @lukerogers , I completely missed this and may have actually fixed in another PR. It does seem like your implementation may better though, so if you're willing to rebase, I think we can get your contribution in.

PR #887, opened 2022-07-22
[stable/fairwinds-insights] Add openapi hpa template

Azahorscak:
**Why This PR?**
Adds an hpa template so hpa's are created when it is enabled in the values file.

Fixes #

**Changes**

*adds open-api hpa
*sets open-api default hpa values to values that won't cause pdb to block rolling node updates 

**Checklist:**

### DIFF
--- a/stable/fairwinds-insights/values.yaml
+++ b/stable/fairwinds-insights/values.yaml
@@ -208,9 +208,9 @@ openApi:
     # -- Create a horizontal pod autoscaler for the Open API server.
     enabled: false
     # -- Minimum number of replicas for the Open API server.
-    min: 1
+    min: 2
     # -- Maximum number of replicas for the Open API server.
-    max: 2
+    max: 3
     # -- Scaling metrics
     metrics:
     - type: Resource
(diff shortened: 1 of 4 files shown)


stable/fairwinds-insights/values.yaml
```diff
@@ -208,9 +208,9 @@ openApi:
     # -- Create a horizontal pod autoscaler for the Open API server.
     enabled: false
     # -- Minimum number of replicas for the Open API server.
-    min: 1
+    min: 2
     # -- Maximum number of replicas for the Open API server.
-    max: 2
+    max: 3
```
rbren, line 213: what do you think about making this `1/2` by default?

It's just serving API documentation, so kind of a secondary feature.

rbren (approved):

stable/fairwinds-insights/values.yaml
Azahorscak: I believe the issue is that if it's set to 1 at any time, the pdb will cause node rolls to fail by blocking the pod deletion. So I believe we should either remove the pdb completely or increase the hpa's defaults.

rbren: Makes sense!

PR #904, opened 2022-08-10
[stable/rbac-manager]  support image digest

swarren83:
**Why This PR?**
Add support for image digest used by cosign to verify signed images.

**Changes**

* support image digest

**Checklist:**


swarren83:
Typically cosign will using a mutating webhook to cross reference tags with the digest to verify the image the tag is for is signed.  In our use case we have ArgoCD that manages resources and the mutating webhook and ArgoCD fight over the state of the resource.  This requires us to use the sha256 digest versus a tag.

sudermanjr (changes requested):
Thanks for the PR! I have always kind of thought we should just start using digest more frequently everywhere :-D 

I'm thinking it might be good to add a file called `ci/digest-values.yaml` and then put an image.digest in that file, so that we run an e2e test using this new value.

swarren83:
@sudermanjr Thank you for the prompt response.  I have added the suggested file.

stable/rbac-manager/ci/digest-values.yaml
```diff
@@ -0,0 +1,2 @@
+image:
+  digest: sha256testvalue
```
sudermanjr: Linter apparently doesn't like this lack of newline

sudermanjr: Oh, and I think we will need to put a valid sha256 for the end to end tests to pass here.

sudermanjr:
```suggestion
  digest: sha256:a2f3d6fb37a22e2ecdb87d8f4f7984c639fa29955cb7e43400ca32804a5c2a2c

```

swarren83:
> Oh, and I think we will need to put a valid sha256 for the end to end tests to pass here. 

I added the v1.4.0 digest sha and the newline.

sudermanjr (approved):
Thanks for the quick fixes!

swarren83:
@sudemanjr Thank you for the quick updates.  Look forward to submit more PRs in the future.

PR #932, opened 2022-09-05
[stable/polaris] Allow user to override container.securityContext

funkypenguin:
**Why This PR?**

Hey guys, for your consideration, I've refactored the polaris chart, so that the container `securityContext` can be amended / altered in `values.yaml`.

Why?

I need to add a seccomp profile to my deployment, and the way the container `securityContext` is currently hard-coded, this is not possible. If the entire `container.securityContext` is moved into `values.yaml`, then users like me could alter / append to it, but if unaltered, the existing `securityContext` will still apply.

**Changes**

* Move Polaris dashboard deployment container securityContext into `values.yaml`

**Checklist:**

### DIFF
--- a/stable/polaris/Chart.yaml
+++ b/stable/polaris/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Validation of best practices in your Kubernetes clusters
 name: polaris
-version: 5.4.2
+version: 5.3.0
 appVersion: "7.0"
 icon: https://raw.githubusercontent.com/FairwindsOps/polaris/master/pkg/dashboard/assets/favicon-32x32.png
 maintainers:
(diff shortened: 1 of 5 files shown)


stable/polaris/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Validation of best practices in your Kubernetes clusters
 name: polaris
-version: 5.4.2
+version: 5.3.0
```
sudermanjr: I think this should be 5.5.0

sudermanjr (changes requested):
This looks great. Thank you for the contribution! One comment about the version change

stable/polaris/Chart.yaml
funkypenguin: Oops, you're right. I semver-bumped backwards! *facepalm*

sudermanjr (approved):

PR #937, opened 2022-09-13
[stable/insights-agent] bump nova version and add new params

vitorvezani:
**Why This PR?**
Bump nova version that has support for `--helm` and `--containers` flags for `find` command
Requires [Nova 3.4 release](https://github.com/FairwindsOps/nova/pull/154)

Fixes #

**Changes**

* Bump version
* Add new flags to the find command

**Checklist:**

### DIFF
--- a/stable/insights-agent/CHANGELOG.md
+++ b/stable/insights-agent/CHANGELOG.md
@@ -1,5 +1,5 @@
 # Changelog
-## 2.6.10
+## 2.6.11
 * Update `nova` version to `v3.4` 
 * Add new flags to `nova find` command 
 
--- a/stable/insights-agent/CHANGELOG.md
+++ b/stable/insights-agent/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 ## 2.6.11
 * Update `nova` version to `v3.4` 
-* Add new flags to `nova find` command 
+* Use `nova find --helm --containers` to run `nova` - This way, both outdated/deprecated charts and containers will be reported to the output file
 
 ## 2.6.10
 Update insights-admission dependency - reattempt of 2.6.9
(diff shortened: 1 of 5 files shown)


stable/insights-agent/templates/nova/cronjob.yaml
```diff
@@ -27,6 +27,8 @@ spec:
             command:
               - /nova
               - find
+              - -- helm
```
vitorvezani: @sudermanjr, @rbren - should these flags be added via config file? Would the below work?
```
nova:
  ...
  config:
    context: ""
    desired-versions: {}
    include-all: true
    wide: true
    output-file: "/output/nova.json"
    helm: true
    containers: true
```

rbren: I'm not sure on this one...I think flags are fine though.

stable/insights-agent/templates/nova/cronjob.yaml
```diff
@@ -27,6 +27,8 @@ spec:
...
+              - -- containers
```
rbren:
```suggestion
              - --helm
              - --containers
```

stable/insights-agent/templates/nova/cronjob.yaml
rbren: Flags are probably better for these, since we don't really want people to override them.

vitorvezani: I wonder if `.Config.configLocation` will not override those flags

rbren (approved):

stable/insights-agent/CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 # Changelog
+## 2.6.11
+* Update `nova` version to `v3.4` 
+* Add new flags to `nova find` command
```
ivanfetch-wt: An non-blocking nit: mentioning what the new flag is/does will help the reader know how nova may behave differently.

ivanfetch-wt (approved):

stable/insights-agent/CHANGELOG.md
vitorvezani: done

Issue #911, opened 2022-08-18
Upgrade to vertical-pod-autoscaler-0.11.0


tanvp112:
Is there a plan to support the latest VPA release?

sudermanjr:
Yes, we definitely plan to at some point. Would love to see a community contribution for this!

PR #960, opened 2022-10-09
[stable/vpa] Update vpa to 0.11.0

dixneuf19:
**Why This PR?**

Update the VPA to the 0.11.0 version. The 0.12.0 has be released since but adds breaking changes so it should be releases in another PR.

The only relevant diff from the `v0.11.0` is 

> Potentially breaking change:
> Added validation - CPU values will be accepted only with resolution of 1 mCPU, memory with resolution of 1 b (https://github.com/kubernetes/autoscaler/pull/4798).

I don't think it a _true_ breaking change, so I did not bump the major version of the Chart, but this can be discussed!

I have run locally the official [VerticalPodAutoscaler repo](https://github.com/kubernetes/autoscaler) script for any "manifest" differences but only the image version changes. The test are still passing

**Checklist:**

### DIFF
--- a/stable/vpa/Chart.yaml
+++ b/stable/vpa/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: vpa
 description: A Helm chart for Kubernetes Vertical Pod Autoscaler
 type: application
-version: 1.4.0
-appVersion: 0.10.0
+version: 1.5.0
+appVersion: 0.11.0
 maintainers:
   - name: sudermanjr


stable/vpa/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: vpa
 description: A Helm chart for Kubernetes Vertical Pod Autoscaler
 type: application
-version: 1.4.0
-appVersion: 0.10.0
+version: 1.4.1
```
sudermanjr: Since this is a minor version change in the underlying project (VPA), I would prefer to make this a minor version change here as well.

sudermanjr (changes requested):
Thanks for the thorough investigation and the PR! One minor request inline.

stable/vpa/Chart.yaml
dixneuf19: I would say that such a change in ["0 sem versioning"](https://0ver.org/) is a major change, but I don't want to trigger a major upgrade with no real breaking change. Anyway, I'm fine with using a minor version here!

I update my commit to release the `1.5.0` instead

sudermanjr: Great, thanks!

sudermanjr (approved):

dixneuf19:
Thanks for the quick review and merge! While that wasn't my main objective, it could be nice that it counts as an Hacktoborfest contribution ( :drooling_face: *craving for free Tshirts*).

If it's OK for you (maybe you wan't to limit contributions on this repo), could you please _add the `hacktoberfest-accepted` label to this PR_ to make it count @sudermanjr ?

And BTW i think you can close #911 , and maybe opening one for the `v0.12.0` (I may look into it if I have some free time).

PR #964, opened 2022-10-11
[stable/fairwinds-insights] Add proper resource requests and limits option for overprovisioning

katiekeel:
**Why This PR?**
The overprovisioning deployment doesn't have a resource requests and limits setup that conforms to K8s norms.

Fixes #

**Changes**

*Adding in proper requests and limits syntax to the chart template.
*Adding in proper requests and limits defaults to the `values.yaml`.

**Checklist:**

* [ x ] I have included the name of the chart in the title of this PR in square brackets i.e. `[stable/goldilocks]`.
* [ x ] I have updated the chart version in `Chart.yaml` following Semantic Versioning.
* [ x ] Any new values are backwards compatible and/or have sensible default.
* [ x ] Any new values have been added to the README for the Chart, or `helm-docs --sort-values-order=file` has been run for the charts that support it.


rbren (approved):

rbren (commented):
I think maybe we should actually just keep the current `values` schema, and always set requests==limits.

Since this pod never actually gets used, there shouldn't be any real resource usage

katiekeel:
> I think maybe we should actually just keep the current `values` schema, and always set requests==limits.
> 
> Since this pod never actually gets used, there shouldn't be any real resource usage

@rbren just so I'm sure what you mean by "keep the current `values` schema" - you're saying we should change the chart template file like I have here, but not edit the defaults in the `values.yaml` file?

rbren:
Right--I'm suggesting that requests should always be equal to limits. So we only need one memory/CPU option in the values.yaml

rbren (approved):

rbren (approved):
Thanks Katie!

sudermanjr (approved):

jdesouza (approved):

PR #1001, opened 2022-11-01
[stable/goldilocks] actually bump image tag

sudermanjr:
**Why This PR?**
actually bump the tag

**Checklist:**

### DIFF
--- a/stable/goldilocks/values.yaml
+++ b/stable/goldilocks/values.yaml
@@ -17,7 +17,7 @@ image:
   # image.repository -- Repository for the goldilocks image
   repository: us-docker.pkg.dev/fairwinds-ops/oss/goldilocks
   # image.tag -- The goldilocks image tag to use
-  tag: v4.5.0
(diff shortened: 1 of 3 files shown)


stable/goldilocks/values.yaml
```diff
@@ -17,7 +17,7 @@ image:
   # image.repository -- Repository for the goldilocks image
   repository: us-docker.pkg.dev/fairwinds-ops/oss/goldilocks
   # image.tag -- The goldilocks image tag to use
-  tag: v4.5.0
+  tag: v4.5.1
```
rbren, line 20: do you want to switch to using `appVersion`? And maybe unpin patch?

rbren (approved):

stable/goldilocks/values.yaml
sudermanjr: at some point in the future would make sense. I don't have the time at the moment

PR #1002, opened 2022-11-03
[stable/rbac-manager] Add deploymentLabels on the Deployment resource.

thib-mary:
**Why This PR?**
Add deploymentLables on the Deployment resource.

**Changes**
* deploymentLables added to deployment.yaml metadata.

**Checklist:**


sudermanjr (changes requested):
Adding podLabels to the deployment object seems inappropriate given the value name. I think it would be preferable to instead add either a deploymentLables value, or add a globalLabels that would label all objects created by the chart (not just deployments)

stable/rbac-manager/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: rbac-manager
-version: 1.13.1
+version: 1.14.1
```
sudermanjr:
```suggestion
version: 1.14.0
```

sudermanjr (changes requested):
Small nit on version. Otherwise looks great! Thanks!

sudermanjr (approved):
Thanks!

PR #1007, opened 2022-11-08
[stable/rbac-manager] Make securityContext configurable

### LINKED ISSUE #1006: [rbac-manager] Make securityContext for pod and container configurable
Using the [Pod Security Standard "restricted" level](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted) requires certain securityContext settings to be set. While the container of rbac-manager actually already fulfills most of the requirements (uid has to be explicitly set if the container image uses a named user) the mandatory seccompProfile is missing and since the securityContext for neither pod nor container is configurable this cannot be amended trivially.

Ideally the current values of the container's `securityContext` would be moved to values.yaml and additionally `podSecurityContext` be made available via values.yaml, too. This would mirror the behaviour of default charts created by `helm create <chart name>` and be fully backwards compatible.

One could probably use kustomize to modify the deployment post-rendering. Since PSS/PSA are default features it will likely be desirable to have a more straightforward solution though.

**Additional context**
N/A

manzsolutions-lpr:
**Why This PR?**
Using the [Pod Security Standard "restricted" level](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted) requires certain securityContext settings to be set. While the container of rbac-manager actually already fulfills most of the requirements (uid has to be explicitly set if the container image uses a named user) the mandatory seccompProfile is missing and since the securityContext for neither pod nor container is configurable this cannot be amended trivially. This PR makes securityContext for both the pod and container fully configurable via values.yaml.

Fixes #1006

**Changes**

* Move container securityContext values to values.yaml as defaults for securityContext
* Add podSecurityContext in values.yaml which is rendered to securityContext on pod level

**Checklist:**

### DIFF
--- a/stable/rbac-manager/Chart.yaml
+++ b/stable/rbac-manager/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: rbac-manager
-version: 1.14.1
+version: 1.15.0
 appVersion: 1.4.0
 description: A Kubernetes operator that simplifies the management of Role Bindings and Service Accounts.
 icon: https://raw.githubusercontent.com/FairwindsOps/charts/master/stable/rbac-manager/icon.png
--- a/stable/rbac-manager/README.md
+++ b/stable/rbac-manager/README.md
@@ -68,6 +68,8 @@ In the above workflow, an RBAC Definition installed between revision 1 and 2 sho
 | affinity | object | `{}` | Deployment affinity |
 | podAnnotations | object | `{}` | Annotations to apply to the pods |
 | podLabels | object | `{}` | Labels to apply to the pod |
+| podSecurityContext | object | `{}` | securityContext to apply to the whole pod |
+| securityContext | object | `{"allowPrivilegeEscalation":false,"capabilities":{"drop":["ALL"]},"privileged":false,"readOnlyRootFilesystem":true,"runAsNonRoot":true}` | securityContext to apply to the rbac-manager container |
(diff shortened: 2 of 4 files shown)


sudermanjr (approved):
Thanks for the PR! This looks great to me. Just need the CLA signed and we're good to go.

manzsolutions-lpr:
@sudermanjr Wonderful. I’m currently waiting for an official sign-off from the company on the CLA, should not take too long.

manzsolutions-lpr:
Unfortunately the “CLA Assistant” requires privileges far beyond what I would be comfortable with:
github com 2022-11-08 17 36 24 a3e44f43c769

So while I would love to sign the CLA I do not see how without authorizing CLA assistant to that extent and that will not happen. Would you happen to know another way? Or am I misinterpreting something?

sudermanjr:
Great question. I think it should only need that level of access if you are setting it up to run your CLA for your own projects. I don't believe signing it should require it. (it seems like it already has access to your email, which is the only thing it should need for signing)

sudermanjr:
Yeah, if I click on the [sign cla](https://cla-assistant.io/FairwindsOps/charts?pullRequest=1007) link it only asks for email once I sign in

manzsolutions-lpr:
I found the issue. I block cookies by default and cla-assistant.io was not whitelisted yet. But instead of throwing an error the callback instead redirected me back to the authorisation page asking for full admin privileges. :-/

PR #1061, opened 2023-01-10
Fix #1059: additionalExemptions

### LINKED ISSUE #1059: [Polaris Helm Chart]: Add optional element additionalExemptions
Polaris Helm Chart renders all configuration elements from a single field `config`. It is thereby not possible to add additional exemptions without replicating all the configurations into each and every usage of the helm chart.

For example, when deploying polaris to "cluster A" and to "cluster B" then we would like the same default exemptions for the two clusters, however, we would also like to add cluster specific exemptions.

**Solution**
Add optional element `additionalExemptions` to the Polaris Helm Chart
Thereby we can create a wrapper Helm Chart that contains the default configuration.
Example of usage:
```
apiVersion: v2
name: polaris-bar
description: A Helm chart for Fairwinds Polaris with specific defaults
version: 0.0.1
keywords: 
  - monitoring
  - security
dependencies:
  - name: polaris
    version: "5.6.0"
```
with default exemptions in the `values.yaml` file in that chart.

"Cluster A" can then use the Helm Chart `polaris-bar` and add extra exemptions:
```
polaris:
  additionalExemptions:
    - namespace: foo
      containerName:
        - hello
      rules:
        - privilegeEscalationAllowed
```

ErikLundJensen:
**Why This PR?**
Support default exemptions and additional exemptions
Fixes #1059 

**Changes**

* The merge-values.yaml is just for testing the chart
* Helm merge function is not used as it does not merge lists. The implementation therefore uses a range loop over the `config` element.

**Checklist:**


sudermanjr:
Hi! Thanks for the contribution! 

You'll need to bump the chart version in the Chart.yaml in order to pass the linter.

stable/polaris/templates/configmap.yaml
```diff
@@ -1,16 +1,20 @@
...
-    {{- toYaml . | nindent 4 }}
-{{- end }}
-{{- end }}
+    {{- range $key, $value := .Values.config }}
+    {{ $key }}:
+    {{- toYaml $value | nindent 6 }}
+    {{- if and (eq $key "exemptions") ($.Values.additionalExemptions) }}
+    {{- toYaml $.Values.additionalExemptions | nindent 6 }}
+    {{- end }}
+    {{- end }}
```
rbren, line 19: This seems a little dangerous to me. If we change the structure of the config (e.g. so `additionalExemptions` isn't last), this will break.

I wonder if we can use a sprig function to combine the two safely

http://masterminds.github.io/sprig/dicts.html

ErikLundJensen: Dicts does not handle arrays, nor merge of arrays.
The order of the elements in config does not matter in this implementation. The additionalExemptions are only added if the key is `exemptions`.
If we change the format of the exemption elements in config then the additionalExemptions must follow the same format.

ErikLundJensen:
I would like to reopen this PR. It was never merged - maybe due to version in Chart.yaml which has been updated.
My fork has been rebased, however, new commits are not process on the closed PR.

rbren:
@ErikLundJensen looks like there's a rebase issue--mind taking a look? Could be easier to just create a new PR too.

Thanks again for the ping!

sudermanjr (changes requested):
~Looks like this needs a rebase.~ Didn't see Robert's comment.

ErikLundJensen:
Right, the rebasing looks wrong. I'll create a new PR (next week).

ErikLundJensen:
Replaced by #1177

PR #1110, opened 2023-03-01
[stable/goldilocks] Set enableArgoproj to false by default

### LINKED ISSUE #1109: Goldilocks RBAC: enableArgoproj
Hi,

When installing Goldilocks, the enableArgoproj value is set to `true` by default.

As people does not have Argo installed in their cluster, I suggest set it to `false`.

rverchere:
**Why This PR?**

Argo being not used by default, disable RBAC related to it.

Fixes #1109

**Changes**

* Set enableArgoproj to false by default
**Checklist:**

### DIFF
--- a/stable/goldilocks/Chart.yaml
+++ b/stable/goldilocks/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 appVersion: "v4.6.3"
-version: 6.5.1
+version: 6.5.2
 description: |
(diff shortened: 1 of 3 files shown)


sudermanjr:
Since this is a non backward compatible change for users, we will need to bump the major version of the chart and note the change in the readme.

rverchere:
> Since this is a non backward compatible change for users, we will need to bump the major version of the chart and note the change in the readme.

Would you like me to make these changes?

PR #1120, opened 2023-03-15
[stable/insights-admission]: Add ability to use a custom SSL certificate when communicating with the Insights API

ivanfetch-wt:
**Why This PR?**
Add the ability to use a custom SSL certificate when communicating with the Insights API. THIs is useful for self-hosted Insights that use a self-signed certificate or certificate authority on the API Kubernetes Ingress.

These new chart values are also used by an [identical feature for the insights-agent chart](https://github.com/FairwindsOps/charts/pull/1121).

**Checklist:**

* [x] I have included the name of the chart in the title of this PR in square brackets i.e. `[stable/goldilocks]`.
* [x] I have updated the chart version in `Chart.yaml` following Semantic Versioning.
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] Any new values have been added to the README for the Chart, or `helm-docs --sort-values-order=file` has been run for the charts that support it.

PR #1121, opened 2023-03-15
[stable/insights-agent]: FWI-3806: Add ability to use a custom SSL certificate when communicating with the Insights API

ivanfetch-wt:
Note this pR requires [a related insights-agent PR](https://github.com/FairwindsOps/charts/pull/1120) to be merged before the tests here will pass.

**Why This PR?**
Add the ability to use a custom SSL certificate when communicating with the Insights API. THIs is useful for self-hosted Insights that use a self-signed certificate or certificate authority on the API Kubernetes Ingress.

These new chart values are also used by an [identical feature for the insights-admission chart](https://github.com/FairwindsOps/charts/pull/1120).

This change optionally allows mounting a volume from a user-supplied Kubernetes secret, which contains a certificate file. This certificate file is included in the certificates used to validate HTTPS communication, and is only included for containers that talk to the Insights API. A few important notes about how this works:
* The environment variable `SSL_CERT_FILE` is set to the path of the certificate file.
* Go binaries use the file specified by `SSL_CERT_FILE` for certificate validation *in addition to* existing certificate authorities that are part of container images such as via the `openssl-ca-certificates` alpine package. This allows 3rd party Go binaries such as trivy to successfully validate other HTTPS connections, such as downloading the trivy data-file.
* The `curl` command will *only* use the file specified by `SSL_CERT_FILE` for certificate validation. For this reason, the Insights fleet installation Kubernetes job does not set the `SSL_CERT_FILE` environment variable until it has used curl to perform other downloads from the Internet.

**Checklist:**

### DIFF
--- a/stable/insights-agent/templates/_uploader.yaml
+++ b/stable/insights-agent/templates/_uploader.yaml
@@ -48,16 +48,12 @@
   {{- toYaml . | nindent 2 }}
   {{- end }}
   {{ include "proxy-env-spec" . | indent 2 | trim }}
-  {{- if and .Values.global.sslCertFile.secretName .Values.global.sslCertFile.secretKey }}
   {{ include "ssl-cert-file-env-spec" . | indent 2 | trim }}
-  {{- end }}
 
   volumeMounts:
   - name: output
     mountPath: /output
-{{- if and .Values.global.sslCertFile.secretName .Values.global.sslCertFile.secretKey }}
 {{ include "ssl-cert-file-volumemount-spec" . | indent 2 }}
-{{- end }}
   resources:
     {{- toYaml .Values.uploader.resources | nindent 4 }}
   securityContext:
--- a/stable/insights-agent/templates/_uploader.yaml
+++ b/stable/insights-agent/templates/_uploader.yaml
@@ -48,9 +48,16 @@
   {{- toYaml . | nindent 2 }}
   {{- end }}
   {{ include "proxy-env-spec" . | indent 2 | trim }}
+  {{- if and .Values.global.sslCertFile.secretName .Values.global.sslCertFile.secretKey }}
+  {{ include "ssl-cert-file-env-spec" . | indent 2 | trim }}
+  {{- end }}
+
   volumeMounts:
   - name: output
     mountPath: /output
+{{- if and .Values.global.sslCertFile.secretName .Values.global.sslCertFile.secretKey }}
+{{ include "ssl-cert-file-volumemount-spec" . | indent 2 }}
+{{- end }}
   resources:
     {{- toYaml .Values.uploader.resources | nindent 4 }}
   securityContext:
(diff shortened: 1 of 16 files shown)


stable/insights-agent/templates/_uploader.yaml
```diff
@@ -48,9 +48,16 @@
...
   {{- end }}
   {{ include "proxy-env-spec" . | indent 2 | trim }}
+  {{- if and .Values.global.sslCertFile.secretName .Values.global.sslCertFile.secretKey }}
+  {{ include "ssl-cert-file-env-spec" . | indent 2 | trim }}
+  {{- end }}
+
   volumeMounts:
   - name: output
     mountPath: /output
+{{- if and .Values.global.sslCertFile.secretName .Values.global.sslCertFile.secretKey }}
```
rbren: it looks like this `if` and the one above are duplicated in side the `ssl-cert-file-volumemount-spec` definition

stable/insights-agent/templates/kube-bench/cronjob.yaml
```diff
@@ -17,8 +17,9 @@ spec:
         {{ include "job-spec-metadata" . | nindent 8 | trim }}
         spec:
           {{ include "job-template-spec" . | indent 10 | trim }}
-          {{- if or (eq (index .Values "kube-bench" "mode") "daemonsetMaster") (eq (index .Values "kube-bench" "mode") "daemonset") }}
           volumes:
+          {{ include "ssl-cert-file-volume-spec" . | indent 10 }}
+          {{- if or (eq (index .Values "kube-bench" "mode") "daemonsetMaster") (eq (index .Values "kube-bench" "mode") "daemonset") }}
```
rbren, line 22: It looks like this line moved into the `volumes` section and out of the `spec` section.

ivanfetch-wt: The primary change is that I moved the `volumes:` above this existing if statement, so I could also include the new cert-related volume. I'll further explain below, but definitely let me know if you still feel this moved something from `spec` to `volumes`.

that existing kube-bench if statement adds a volume definition, and goes on to add a container, when kube-bench is run in daemonset mode.

The content of the if statement includes `containers:`, as does the `kubebench.container.spec` define which is used in the else statement below this if.

stable/insights-agent/templates/_uploader.yaml
ivanfetch-wt: Ah yes - thanks, I missed that one after I moved the if inside of the defines.

stable/insights-agent/templates/kube-bench/cronjob.yaml
rbren: got it!

rbren (approved):

PR #1122, opened 2023-03-17
[incubator/vault-token-injector] Allow using a cronjob

sudermanjr:
**Why This PR?**
Sometimes, the best way to run a regular task is a cronjob

**Changes**

* Add the ability to run as a cronjob
* reorganize a few things
* Update versions

**Checklist:**

### DIFF
--- a/incubator/vault-token-injector/values.yaml
+++ b/incubator/vault-token-injector/values.yaml
@@ -1,5 +1,38 @@
-# -- We currently only support a single instance
-replicaCount: 1
+deployment:
+  # -- If true, will be run as a deployment. Technically both cronjob and deployment can be enabled at the same time, but why?
+  enabled: true
+  # -- We currently only recommend a single instance
+  replicaCount: 1
+  probes:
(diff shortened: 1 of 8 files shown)


incubator/vault-token-injector/values.yaml
```diff
@@ -25,7 +48,7 @@ config:
     env_variable: VAULT_TOKEN
 
 metrics:
-  # -- If true, a prometheus endpoint will be enabled on port 4329
+  # -- If true, a prometheus endpoint will be enabled on port 4329. This only applies to the deployment
```
rbren: Maybe just put this under `deployment` then? Not a blocker.

rbren (approved):

incubator/vault-token-injector/values.yaml
sudermanjr: I thought about that, and originally was trying not to introduce breaking changes, but then I did anyway. Will update.

Issue #1161, opened 2023-04-26
[vpa] Inadequate template condition checks for some templates 


awsitcloudpro:
### What happened?

Some of the VPA chart templates have inadequate checks for resource provisioning. For example:
- Even when `.Values.admissionController.enabled` is `false`, if `admissionController.cleanupOnDelete.enabled` is `true`, then the resources in the template admission-controller-cleanup.yaml are still deployed.
- Similarly, PDB templates such as recommender-pdb.yaml are not checking if the relevant VPA component is enabled or not

This is similar to issue #722 - looks like the bug got reintroduced.

### What did you expect to happen?

If a component is disabled, all resources related to that component should not be provisioned. All templates related to a component should contain a check for the component's `enabled` flag - e.g. `{{- if .Values.admissionController.enabled }}` should exist in admission-controller-cleanup.yaml.

### How can we reproduce this?

Inspect the code of templates mentioned above or deploy the current stable version of the VPA chart and observe what resources are provisioned.

### Version

vpa - 1.7.2

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

sudermanjr:
In general, I feel like the way we're handling the admission configurations, certificates, etc. could really use a revamp. There are better patterns out there for this out there.

Issue #854, opened 2022-07-01
VPA Admission Controller Hooks Multiarch Image


jordanfelle:
**Is your feature request related to a problem? Please describe.**
Running the VPA on an ARM64 only cluster fails if Admission Controller is enabled

**Describe the solution you'd like**
Switch to a multiarch image that will let the helm chart work in arm64 only clusters

sudermanjr:
Is there a publicly available multi-arch image available? I would be happy to accept a PR to change the image. We don't maintain the VPA project, so I don't control what images are available by default, and maintaining a pipeline to build new ones is a bit out of scope for us at the moment.

jordanfelle:
@sudermanjr Sorry, I should have specified the image that is the problem. The VPA images are multiarch and it looks like the only image that isn't is one that is not part of the official VPA images at https://github.com/FairwindsOps/charts/blob/master/stable/vpa/values.yaml#L135 . Looking at it on quay the layers look like they are setup to be built for mutliarch but only the amd64 was pushed up.

sudermanjr:
Ahhh, I see. Yeah, that's one of our images. I don't see us moving that particular one to be multiarch due to the process being much much older. 

I think we could probably swap that out. We only used it for convenience at the time.

sudermanjr:
## ✅ Pull Request Created

I've created a pull request to address this issue: #1980

### Changes Made
- Updated the `kube-webhook-certgen` image tag from `v20230312-helm-chart-4.5.2-28-g66a760794` to `v1.4.4`
- This version includes proper multiarch support for ARM64 clusters

### Summary
The admission controller's certificate generation hook now uses the latest stable version (v1.4.4) from the Kubernetes ingress-nginx project, which has confirmed multiarch support. This enables the VPA chart to work on ARM64-only clusters when the admission controller is enabled.

---
*This comment was posted by an AI agent (OpenHands).*

PR #1191, opened 2023-05-17
[stable/vpa] Refactor webhook and cert handling

schoeppi5:
**Why This PR?**
The certification creation process was unnecessary complex using openssl. This PR aims to simplify certificate creation using a tool specifically designed for this process.
Since the MutatingWebhookConfiguration wasn't actually managed Helm, a workaround using cleanup jobs was needed.
The image used for testing wasn't configurable.

Fixes #

* Removed redundant namespace declarations
* Use release name for all resources
* Further limited RBAC configuration to specific resources

**Changes**

* Reworked certificate creation process
* Create MutatingWebhookConfiguration using helm. Allows for templating
* Configurable testing image. Switched to bitnami/kubectl

**Checklist:**

<sub>Philipp Schöppner <[philipp.schoeppner@mercedes-benz.com](mailto:philipp.schoeppner@mercedes-benz.com)>, Mercedes-Benz Tech Innovation GmbH ([Provider Information](https://github.com/mercedes-benz/daimler-foss/blob/master/PROVIDER_INFORMATION.md))</sub>


schoeppi5:
Would my changes be considered a major update, if a direct upgrade path (as tested by the e2e tests) is not possible (as described in the readme)?

sudermanjr (changes requested):
Thanks for the contribution! I haven't had a chance to look through in detail, but first glance looks really great.

If this can't be done as a straight upgrade, then yes, this will require a major version update and instructions on how to upgrade in the README. 

Looking at the tests, the error we're getting is:

```
Error: UPGRADE FAILED: template: vpa/templates/webhooks/mutating.yaml:9:18: executing "vpa/templates/webhooks/mutating.yaml" at <.Values.admissionController.mutatingWebhookConfiguration.annotations>: nil pointer evaluating interface {}.annotations
```

You might be able to mitigate this and allow for a minor version bump with a well-placed {{if}} in a template (I've done this in other changes before)

schoeppi5:
I think I was able to mitigate all breaking changes, so that this can remain a minor version upgrade, rather than a breaking major version. Please let me know, if you need anything additional from me!

schoeppi5:
This test failure is expected, since the e2e tests don't install the metrics server (afaik). What can I do, to resolve this issue?

schoeppi5:
I just had a look through the open issues for this chart and #854 can be closed by this PR. #1161 would require a few more changes to fully satisfy.

Edit: Clarification on #854: Both the bitnami/kubectl image as well as the kube-webhook-certgen image are available for at least amd64 and arm64

schoeppi5:
My latest commit also fully closes #1161.

stable/vpa/.gitignore
```diff
@@ -0,0 +1,2 @@
+charts/
+Chart.lock
```
sudermanjr: I don't think it's best practice to exclude this, is it?

stable/vpa/README.md
```diff
@@ -135,3 +175,7 @@ recommender:
 | admissionController.tolerations | list | `[]` |  |
 | admissionController.affinity | object | `{}` |  |
 | tests.securityContext | object | `{"allowPrivilegeEscalation":false,"capabilities":{"drop":["ALL"]},"readOnlyRootFilesystem":true,"runAsNonRoot":true,"runAsUser":10324}` | The security context for the containers run as helm hook tests |
+| tests.image.repository | string | `"bitnami/kubectl"` | An image used for testing containing bash, cat and kubectl |
+| tests.image.tag | string | `""` | An image tag for the tests image |
+| tests.image.pullPolicy | string | `"Always"` | The pull policy for the tests image. |
+| metrics-server.enabled | bool | `true` |  |
```
sudermanjr: This either needs to default to "false" with a change to ci/test-values, or be moved to a pre-test script for installation during tests. Most environments will not need this, and it will break existing users of Goldilocks, which I would like to avoid.

https://github.com/FairwindsOps/charts/blob/master/scripts/pre-test-script-runner.sh

sudermanjr (changes requested):

stable/vpa/.gitignore
schoeppi5: I oriented myself on the other examples in this repository.
I will change that, thanks for the comment.

stable/vpa/README.md
schoeppi5: I would opt for keeping it as a dependency and disabling it by default. I think, it clarifies, that VPA has the metrics-server as a dependency.
Thank you for your comment, as I was not aware of the implications of this change.

stable/vpa/README.md
```diff
@@ -2,16 +2,16 @@
...
 
 There are a few tests included with this chart that can help debug why your installation of VPA isn't working as expected. You can run `helm test -n <Release Namespace> <Release Name>` to run them.
 
-* `checkpoint-crd-available` - Checks for the verticalpodautoscalercheckpoints CRD
-* `crd-available` - Checks for the verticalpodautoscalers CRD
-* `metrics-api-available` - Checks to make sure that the metrics api endpoint is available. If it's not, install [metrics-server](https://github.com/kubernetes-sigs/metrics-server) in your cluster
+* `crds-available` - Checks for both the _verticalpodautoscalers_ and _verticalpodautoscalercheckpoints_ CRDs
+* `metrics-api-available` - Checks to make sure that the metrics API endpoint is available. If it's not, install [metrics-server](https://github.com/kubernetes-sigs/metrics-server) in your cluster.
 * `create-vpa` - A simple check to make sure that VPA objects can be created in your cluster. Does not check for functionality of that VPA.
++ `webhook-configuration` - Checks, that both the service and the CA bundle in the MutatingWebhookconfiguration are configured correctly.
```
transient1:
```suggestion
+ `webhook-configuration` - Checks that both the service and the CA bundle in the MutatingWebhookconfiguration are configured correctly.
```

stable/vpa/README.md
```diff
@@ -21,15 +21,56 @@ There are three primary components to the Vertical Pod Autoscaler that can be en
...
 For more details, please see the values below, and the vertical pod autosclaer documentation.
 
-## *BREAKING* Upgrading from v0.x.x to v1.x.x
+## *BREAKING* Upgrading from <= v1.7.x to 2.0.0
+
+### Certificate generation
 
-In the previus version, when admissionController.cleanupOnDelete flag passed to true, mutatingwebhookconfiguration and the tls secret for the admission controller is removed. There was no chance to pass any image information to start remove process. Now, it could be passed custom image by version 1.0.0.
+The certificate creation process was changed from using OpenSSL to [kube-webhook-certgen](https://github.com/kubernetes/ingress-nginx/tree/main/images/kube-webhook-certgen) to simplify the process.
+It still uses the same configuration keys (.Values.admissionController.certGen), which makes it impossible to reuse the values from a previous install.
```
transient1, line 33:
```suggestion
It still uses the same configuration keys (.Values.admissionController.certGen), which makes it possible to reuse the values from a previous install.
```
Is this supposed to mean that you _can_ reuse?

stable/vpa/README.md
```diff
@@ -21,15 +21,56 @@ There are three primary components to the Vertical Pod Autoscaler that can be en
...
+  --set "admissionController.certGen.image.tag=v20230312-helm-chart-4.5.2-28-g66a760794"
 ```
+
+The new process is incompatible with the old secrets layout. To mitigate this, the secret was renamed to (by default) `<release name>-tls-certs` and can now also be customized.
+
+All other changes are implemented in a non breaking fashion.
+
+### MutatingWebhookconfiguration
+
+Previously, the webhook creation was handled by the admission controller itself. This had the downside, that Helm is not in control of the resource and therefore required the cleanupOnDelete job.
```
transient1:
```suggestion
Previously, the webhook creation was handled by the admission controller itself. This had the downside that Helm is not in control of the resource and therefore required the cleanupOnDelete job.
```

stable/vpa/README.md
```diff
@@ -21,15 +21,56 @@ There are three primary components to the Vertical Pod Autoscaler that can be en
...
+
+Also, the `cleanupOnDelete` configuration is obsolete.
+
+### Admission controller
+
+The admission controller is enabled by default.
+
+## *BREAKING* Upgrading from v0.x.x to v1.x.x
+
+In the previus version, when admissionController.cleanupOnDelete flag passed to true, MutatingWebhookconfiguration and the tls secret for the admission controller is removed. There was no chance to pass any image information to start remove process. Now, it could be passed custom image by version 1.0.0.
```
transient1:
```suggestion
In the previous version, when the admissionController.cleanupOnDelete flag was set to true, MutatingWebhookconfiguration and the tls secret for the admission controller were removed. There was no chance to pass any image information to start remove process. Now, it could be passed custom image by version 1.0.0.
```

stable/vpa/README.md.gotmpl
```diff
@@ -21,15 +21,56 @@ There are three primary components to the Vertical Pod Autoscaler that can be en
...
 
 For more details, please see the values below, and the vertical pod autosclaer documentation.
 
-## *BREAKING* Upgrading from v0.x.x to v1.x.x
+## *BREAKING* Upgrading from <= v1.7.x to 2.0.0
+
+### Certificate generation
+
+The certificate creation process was changed from using OpenSSL to [kube-webhook-certgen](https://github.com/kubernetes/ingress-nginx/tree/main/images/kube-webhook-certgen) to simplify the process.
+It still uses the same configuration keys (.Values.admissionController.certGen), which makes it impossible to reuse the values from a previous install.
```
transient1, line 33:
```suggestion
It still uses the same configuration keys (.Values.admissionController.certGen), which makes it possible to reuse the values from a previous install.
```
Same as above. Is it possible because of the same configuration key?

stable/vpa/README.md.gotmpl
```diff
@@ -21,15 +21,56 @@ There are three primary components to the Vertical Pod Autoscaler that can be en
...
+  --set "admissionController.certGen.image.tag=v20230312-helm-chart-4.5.2-28-g66a760794"
 ```
+
+The new process is incompatible with the old secrets layout. To mitigate this, the secret was renamed to (by default) `<release name>-tls-certs` and can now also be customized.
+
+All other changes are implemented in a non breaking fashion.
+
+### MutatingWebhookconfiguration
+
+Previously, the webhook creation was handled by the admission controller itself. This had the downside, that Helm is not in control of the resource and therefore required the cleanupOnDelete job.
```
transient1:
```suggestion
Previously, the webhook creation was handled by the admission controller itself. This had the downside that Helm is not in control of the resource and therefore required the cleanupOnDelete job.
```

stable/vpa/README.md.gotmpl
```diff
@@ -21,15 +21,56 @@ There are three primary components to the Vertical Pod Autoscaler that can be en
...
+
+Also, the `cleanupOnDelete` configuration is obsolete.
+
+### Admission controller
+
+The admission controller is enabled by default.
+
+## *BREAKING* Upgrading from v0.x.x to v1.x.x
+
+In the previus version, when admissionController.cleanupOnDelete flag passed to true, MutatingWebhookconfiguration and the tls secret for the admission controller is removed. There was no chance to pass any image information to start remove process. Now, it could be passed custom image by version 1.0.0.
```
transient1:
```suggestion
In the previous version, when admissionController.cleanupOnDelete flag was set to true, MutatingWebhookconfiguration and the tls secret for the admission controller were removed. There was no chance to pass any image information to start remove process. Now, it could be passed custom image by version 1.0.0.
```

stable/vpa/templates/webhooks/jobs/certgen-patch.yaml
```diff
@@ -0,0 +1,54 @@
...
+kind: Job
+metadata:
+  name:  {{ include "vpa.fullname" . }}-admission-certgen-patch
+  annotations:
+    "helm.sh/hook": post-install,post-upgrade
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+  labels:
+    app.kubernetes.io/component: admission-certgen
+    {{- include "vpa.labels" . | nindent 4 }}
+spec:
```
rbren, line 12: We should add a `ttlSecondsAfterFinished` here and for other jobs. Maybe `300` (5 minutes)?

I generally prefer this to the helm-hook strategy, since hooks can cause so many problems (e.g. with GitOps strategies)

sudermanjr (changes requested):
A few small tweaks from reviewers, and a rebase and we should be good to merge.

stable/vpa/README.md
schoeppi5: Thank you for your comment. Maybe I should clarify this statement.

The issue here is as follows:

If you reuse the values of a previous install (`helm upgrade --reuse-values ...`), the mentioned configuration keys (`.Values.admissionController.certGen.repository, .tag, etc) contain the information on the previously used image for this job, which makes them incompatible with the new job configuration (arguments passed as part of the pod spec mean nothing to the ci image).

Therefore, a straightforward update from the previous version is not an option (impossible), which leads to this PR being a major version change

sudermanjr: I think maybe re-wording to say "It uses the same Values from the previous version, which makes it impossible to upgrade directly using the helm `--reuse-values` flag."

sudermanjr (approved):

PR #1220, opened 2023-06-12
FWI-4251 - Add support for readreplica to hubspot cronjob

jslivka:
**Why This PR?**

This is to address the need for using a dedicated read replica when running the hubspot cronjob in a SaaS context. Given that we don't want to negotiate another environment variable in the go database client, and multiple workloads are configured through a singular helm chart for fairwinds-insights, this is a minimally invasive approach to override the `POSTGRES_*` environment variables for only the hubspot cronjob. Once we introduce read replica support for other workloads, I think we should do away with how we currently use the monolithic `_env.yaml` template and also use subcharts for distinct workloads.

Fixes #

**Changes**

* Checks if the current template being rendered is `fairwinds-insights/templates/hubspot-cronjob.yaml` and whether `postgresql.readReplica` is configured. This assumes the presence of new `database`, `username`, `port`, `sslMode` and `host` values. These all refer to a postgresql read replica
* Assumes the presence of a new key `readreplica-password` in the existing `postgresql` secret:
```
- name: POSTGRES_PASSWORD
  valueFrom:
    secretKeyRef:
      name: {{ .Values.postgresql.auth.existingSecret }}
      key: readreplica-password
```

## Template

```bash
helm template . -f values.yaml -f <(cat <<END
options:
  insightsSAASHost: null
postgresql:
  readReplica:
    username: username
    database: postgres
    host: pghost
    port: 5432
    sslMode: require
hubspotCronjob:
  schedules:
    - name: hourly
      interval: 60m
      cron: "0 * * * *"
END
) \
--show-only templates/hubspot-cronjob.yaml
```

```yaml
# Database
- name: POSTGRES_DB
  value: postgres
- name: POSTGRES_USER
  value: username
- name: POSTGRES_PORT
  value: "5432"
- name: POSTGRES_SSL_MODE
  value: require
- name: POSTGRES_HOST
  value: pghost
- name: POSTGRES_PASSWORD
  valueFrom:
    secretKeyRef:
      name: fwinsights-postgresql
      key: readreplica-password
```

**Checklist:**


stable/fairwinds-insights/templates/_env.yaml
```diff
@@ -50,6 +50,23 @@ env:
   value: {{ . | quote }}
 {{- end }}
 # Database
+{{- if and (.Values.hubspotCronjob.readReplica) (eq .Template.Name "fairwinds-insights/templates/hubspot-cronjob.yaml") }}
```
rbren: oh nice! I didn't know you could use `.Template.Name`

stable/fairwinds-insights/values.yaml
```diff
@@ -81,6 +81,13 @@ hubspotCronjob:
       cpu: 80m
       memory: 128Mi
   schedules: []
+  # -- Optional read replica configuration
+  readReplica:
```
rbren: we might want to reuse this in other places--maybe we should put it under `postgres`

jslivka: I thought it may hurt readability to scope it to `postgres`, because if you're reading the chart it's not clear what the read replicas are for. I suppose I could make this change and address that in the README if it works for you

stable/fairwinds-insights/templates/_env.yaml
jslivka: I really didn't want to do it this way at first but doing weird merge operations with maps in sprig is just gross

jslivka:
It would also be helpful to know for the subchart exercise which workloads in this chart communicate with postgres, and of those: the ones that need write access, read access or both

stable/fairwinds-insights/templates/_env.yaml
rbren: Haha I was gonna suggest the merge and then decided against it

stable/fairwinds-insights/values.yaml
rbren: Yeah good point--we could pull it out as a top-level option too. I just don't want to scope it to `hubspot` and then decide it's useful somewhere else

rbren:
Good question on which workloads do what--I _think_ the vast majority do read/write postgres--hubspot is the only one we've identified so far as a good read-replica candidate

rbren (approved):

PR #1263, opened 2023-07-17
FWI-4570 - Add Kyverno plugin to [stable/insights-agent]

jslivka:
**Why This PR?**
This introduces helm configuration for the `kyverno` insights plugin

**Checklist:**

### DIFF
--- a/stable/insights-agent/templates/kyverno/cronjob.yaml
+++ b/stable/insights-agent/templates/kyverno/cronjob.yaml
@@ -1,0 +1,36 @@
+{{- if and .Values.kyverno .Values.kyverno.enabled -}}
+{{- $_ := set . "Label" "kyverno" }}
+{{- $_ := set . "Config" .Values.kyverno }}
+{{- include "cronjob" . }}
+spec:
+  {{ include "cronjob-spec" . | nindent 2 | trim }}
+  jobTemplate:
+    spec:
+      {{ include "job-spec" . | nindent 6 | trim }}
+      template:
+        {{ include "job-spec-metadata" . | nindent 8 | trim }}
+        spec:
+          {{ include "job-template-spec" . | indent 10 | trim }}
+          containers:
+          - {{ include "container-spec" . | indent 12 | trim }}
+            env:
+            - name: FAIRWINDS_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  {{ if .Values.insights.tokenSecretName -}}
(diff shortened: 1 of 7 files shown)


stable/insights-agent/templates/kyverno/cronjob.yaml
```diff
@@ -0,0 +1,36 @@
...
+                  {{ else -}}
+                  name: {{ include "insights-agent.fullname" . }}-token
+                  {{ end -}}
+                  key: token
+            - name: FAIRWINDS_INSIGHTS_HOST
+              value: {{ .Values.insights.host }}
+            - name: FAIRWINDS_ORG
+              value: {{ .Values.insights.organization | quote  }}
+            - name: FAIRWINDS_CLUSTER
+              value: {{ .Values.insights.cluster | quote }}
```
rbren: 🤔 do we need these?

jslivka: I was under the impression that the uploader container needs the host/cluster/org/token or am I reading that wrong? https://github.com/FairwindsOps/charts/blob/effe98e2992a0e5a6482e6e5600087b3ef4adbd2/stable/insights-agent/templates/_uploader.yaml

jslivka: Or am I adding duplication from the template here? https://github.com/FairwindsOps/charts/blob/effe98e2992a0e5a6482e6e5600087b3ef4adbd2/stable/insights-agent/templates/_uploader.yaml

Was odd that the OPA cronjob is set up like that

rbren: ahh OPA needs some of these things because it _downloads_ policies via the API. This setup should be simpler

rbren (approved):

stable/insights-agent/values.yaml
```diff
@@ -575,3 +575,15 @@ falcosecurity:
         minimumpriority: ""
         mutualtls: false
         checkcert: true
+
+kyverno:
+  enabled: false
+  schedule: "rand * * * *"
+  image:
+    repository: quay.io/fairwinds/kyverno
+    tag: "0.1.1"
```
rbren: I'd suggest unpinning the patch version, like we do for other plugins
```suggestion
    tag: "0.1"
```

rbren (approved):

PR #1325, opened 2023-09-18
[stable/vpa] Make an opportunity to extend rbac ClusterRoles

asenyaev:
**Why This PR?**
At the moment, there are no opportunity to add extra rules in `ClusterRole`, and for some use-cases it is a must.

Fixes #

**Changes**

* This PR make an opportunity to extend rbac ClusterRoles

**Checklist:**

### DIFF
--- a/stable/vpa/values.yaml
+++ b/stable/vpa/values.yaml
@@ -16,19 +16,19 @@ rbac:
   # rbac.create -- If true, then rbac resources (ClusterRoles and ClusterRoleBindings) will be created for the selected components.
   # Temporary rbac resources will still be created, to ensure a functioning installation process
   create: true
-  # rbac.clusterRoles -- Extra rbac rules for Clusterroles
-  clusterRoles:
-    # rbac.clusterRoles.vpaActor -- Extra rbac rules for the vpa-actor ClusterRole
+  # rbac.extraRules -- Extra rbac rules for ClusterRoles
+  extraRules:
+    # rbac.extraRules.vpaActor -- Extra rbac rules for the vpa-actor ClusterRole
     vpaActor: []
-    # rbac.clusterRoles.vpaCheckpointActor -- Extra rbac rules for the vpa-checkpoint-actor ClusterRole
+    # rbac.extraRules.vpaCheckpointActor -- Extra rbac rules for the vpa-checkpoint-actor ClusterRole
     vpaCheckpointActor: []
-    # rbac.clusterRoles.vpaEvictioner -- Extra rbac rules for the vpa-evictioner ClusterRole
+    # rbac.extraRules.vpaEvictioner -- Extra rbac rules for the vpa-evictioner ClusterRole
     vpaEvictioner: []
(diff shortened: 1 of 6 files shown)


stable/vpa/values.yaml
```diff
@@ -13,9 +13,23 @@ fullnameOverride: ""
 # podLabels -- Labels to add to all pods
 podLabels: {}
 rbac:
-  # rbac.create -- If true, then rbac resources (clusterroles and clusterrolebindings) will be created for the selected components.
+  # rbac.create -- If true, then rbac resources (ClusterRoles and ClusterRoleBindings) will be created for the selected components.
   # Temporary rbac resources will still be created, to ensure a functioning installation process
   create: true
+  # rbac.clusterRoles -- Extra rbac rules for Clusterroles
+  clusterRoles:
```
sudermanjr: I would prefer to use a more descriptive name, such as `additionalClusterRoles`

sudermanjr (changes requested):
Thanks! 

Mostly LGTM, just a slight change to the naming requested.

sudermanjr:
Also, could you please add some example additional roles to the ci/test-values.yaml file so that we can cover this functionality in tests?

stable/vpa/values.yaml
asenyaev: I think your proposal can confuse a user, because `additionalClusterRoles` can be interpreted as new ClusterRoles, but in this case they already exist.

sudermanjr: to me `rbac.clusterRoles` indicates that you're setting _all_ the clusterroles, not adding extra ones. Do you have a different name in mind that might be more descriptive and avoid the confusion you suggest?

asenyaev: It can be the same as it is in `goldilocks` - [`extraRules`](https://github.com/FairwindsOps/charts/blob/master/stable/goldilocks/values.yaml#L40-L41). What do you think?

asenyaev: Another idea in my head - `additionalClusterRoleRules`. It's a bit long, but as clear as possible (:

asenyaev:
@sudermanjr I've tried to add some breaking info in `README.md`, but `helm-docs` doesn't like it. Should I remove these lines?

sudermanjr:
You can add the info to the readme template. `README.md.gotmpl`

stable/vpa/values.yaml
sudermanjr: Either of those is fine for me. I like being consistent with the Goldilocks chart.

sudermanjr:
@asenyaev apologies, I accidentally clicked the `update branch` button - bad habit.

asenyaev:
@sudermanjr no problem, it's even better tbh. (:

stable/vpa/values.yaml
asenyaev: I've renamed `clusterRoles` value to `extraRules`.

sudermanjr (approved):

PR #1329, opened 2023-09-27
[stable/goldilocks] Turn off resource limits by default

asenyaev:
**Why This PR?**
At the moment, you cannot turn off resources, because it is in values. From time to time you need to define only `requests` and keep the `limits` flexible.

Fixes #

**Changes**

* This PR turns off resources for dashboard and controller by default, and make an opportunity to define `requests` only. 

**Checklist:**

### DIFF
--- a/stable/goldilocks/Chart.yaml
+++ b/stable/goldilocks/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 appVersion: "v4.10.0"
-version: 7.2.0
+version: 7.2.1
 kubeVersion: ">= 1.22.0-0"
 description: |
   A Helm chart for running Fairwinds Goldilocks. See https://github.com/FairwindsOps/goldilocks
--- a/stable/goldilocks/Chart.yaml
+++ b/stable/goldilocks/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 appVersion: "v4.10.0"
-version: 7.2.1
+version: 7.3.0
 kubeVersion: ">= 1.22.0-0"
 description: |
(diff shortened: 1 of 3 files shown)


sudermanjr:
You can turn off resources by setting resources: null in your values file.

sudermanjr:
I would support removing the default limits, but not the default requests.

sudermanjr (changes requested):
I would support removing the default limits, but not the default requests.

asenyaev:
> You can turn off resources by setting resources: null in your values file.

The main idea to support an opportunity to define (or not) requests / limits. I just want to support a [theory](https://home.robusta.dev/blog/stop-using-cpu-limits) to not to use CPU limits, but there is no way to do that.

sudermanjr:
Yes there is, you can set the limits to null in your values

sudermanjr:
https://stackoverflow.com/questions/63766999/unset-remove-default-value-in-helm-values-yaml

asenyaev:
> I would support removing the default limits, but not the default requests.

Basically, helm charts don't have pre-defined resources or at least have resources when the developer wants to help a user with guaranteed resources to start an application.

> Yes there is, you can set the limits to null in your values

Yes, sure, but usually helm charts don't have pre-defined resources and in my opinion the current state is not the best way. However, you are the owner of this repository and have your own vision.

sudermanjr:
I will happily support a PR to remove the limits, but not the requests. I understand that default limits cause issues for some users out of the box.

Setting default requests increases reliability and is a baseline requirement for all of our OSS.

sudermanjr (changes requested):
Also please make a change in default values a minor version bump

sudermanjr (approved):

PR #1345, opened 2023-10-13
[stable/vpa] Update securityContext defaults

bodgit:
**Why This PR?**
We run Kyverno on our clusters which alerts if the various `securityContext` settings aren't explicitly configured. Your Helm chart allows me to set/override these, but before I push this out to more clusters I figured it would be easier if the chart defaults were just updated.

As I'm only using the recommender, (for Goldilocks), I've not touched the `updater` or `admissionController` sections. I'm guessing these would probably be fine to have the same changes applied so can update the PR if desired.

**Changes**

* Update recommender.podSecurityContext defaults to include `seccompProfile`.
* Update recommender.securityContext defaults to set `readOnlyRootFilesystem`, `allowPrivilegeEscalation` & `capabilities`.

**Checklist:**

### DIFF
--- a/stable/vpa/Chart.yaml
+++ b/stable/vpa/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: vpa
 description: A Helm chart for Kubernetes Vertical Pod Autoscaler
 type: application
-version: 4.0.1
+version: 4.1.0
 appVersion: 1.0.0
 maintainers:
   - name: sudermanjr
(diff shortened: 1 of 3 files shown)


sudermanjr (changes requested):
Since this changes default values, I would rather see a minor version change.

Also, if you don't mind keeping it consistent across all three deployments, I would also prefer that. Thanks!

bodgit:
@sudermanjr done

bodgit:
@sudermanjr ping

sudermanjr (approved):

PR #1365, opened 2023-11-08
[stable/vpa] Add admissionController.useHostNetwork

skruppy:
**Why This PR?**
_Using CNI on AWS EKS requires the pod to use host network to be accessible by the control plane._

**Changes**

* Add `admissionController.useHostNetwork` variable (defaults to `false`, so it's backwards compatible)
* Add `admissionController.httpPort` and `admissionController.metricsPort` variable to switch ports if hosts ports are clashing with other pods.

**Checklist:**


skruppy:
It seems like the tests fail [since new values are added and `--reuse-values` is used](https://gitlab.com/gitlab-org/charts/gitlab-agent/-/issues/43#note_1447361539).

Should it hence be marked as a breaking change or piggybacked onto the new major release of #1355?

skruppy:
I just saw that some commits (https://github.com/FairwindsOps/charts/commit/6e6b3d3050c7d42ca00c3f52a27e96bc6afb2cbf) use the `default` function to mitigate the upgrade issue, described in my previous comment, but this contradicts the [contribution guide lines](https://github.com/FairwindsOps/charts/blob/master/CONTRIBUTING.md#configuration). What do you prefer @sudermanjr ?

sudermanjr:
I'm okay with this being a minor change and using the default function, especially since hostNetwork should usually be false for a lot of folks. Another option is to wrap the hostNetwork piece in an if statement, checking if the value exists or not. 

The --reuse-values bit has gotten particularly frustrating as of late - it should be easier to add new values.

sudermanjr (approved):

PR #1425, opened 2024-02-16
[stable/insights-agent] FWI-5698 Right-sizer rebrand

jslivka:
This change replaces the existing `right-sizer` with the new Insights `right-sizer`, which will have additional capabilities for automated right-sizing workloads. This is now a dependency of `right-sizer`, and named `oom-detection` for clarity. Some things to note:

* `right-sizer.oom-detection.enabled: true` must be set to enable what was called right-sizer prior to this change
* `INSIGHTS_TOKEN` is necessary for the VPA recommender to query Insights recommendations, this was accomplished by adding it to the insights agent secret and pulled in with `vpa.recommender.envFromSecret: insights-agent-token`
* There will still be some naming confusion with the actual binaries, as the oom detection is named `right-sizer`, and the VPA controller is named `insights-right-sizer`. We will address this soon.

Helm install command for testing:

```
helm repo update
helm upgrade --install insights-agent . -f values2.yaml \
  --create-namespace \
  --namespace insights-agent \
  --wait \
  --atomic
```

```
# values2.yaml
insights:
...
prometheus-metrics:
  enabled: true
  installPrometheusServer: true
workloads:
  enabled: true
right-sizer:
  enabled: true
```

**Checklist:**


rbren:
Let's call this `right-sizer` instead of `insights-right-sizer`

stable/insights-agent/templates/token-secret.yaml
```diff
@@ -5,4 +5,5 @@ metadata:
   name: {{ include "insights-agent.fullname" . }}-token
 data:
   token: {{ required "You must set base64token or tokenSecretName" .Values.insights.base64token | quote }}
+  INSIGHTS_TOKEN: {{ required "You must set base64token or tokenSecretName" .Values.insights.base64token | quote }}
```
rbren, line 8: Where is this getting used? Is it possible to reuse `token` instead?

rbren (approved):

PR #1501, opened 2024-06-13
[stable/vpa] add securityContext for certgen pods

tibuntu:
**Why This PR?**
This PR adds individual securityContext for the containers belonging to the certgen jobs and renames the current Helm-Value `securityContext` to `podSecurityContext` because it fits the use case better (Also noticed that you use the same naming for both deployments already).

Fixes #
-

**Changes**
-

**Checklist:**

### DIFF
--- a/stable/vpa/Chart.yaml
+++ b/stable/vpa/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: vpa
 description: A Helm chart for Kubernetes Vertical Pod Autoscaler
 type: application
-version: 4.4.6
+version: 4.5.0
 appVersion: 1.0.0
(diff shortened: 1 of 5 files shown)


sudermanjr (commented):
This looks great, my only ask would be to consider adding a security context to the ci/test-values.yaml so that these get templated during the tests

sudermanjr (approved):

tibuntu:
@sudermanjr thanks for the quick review and merge! Could you elaborate why the Goldilocks Helm-Chart isn't using one of the most recent VPA charts? I was a little bit confused why my change wasn't present in the recent Goldilock Helm-Chart.

PR #1510, opened 2024-07-17
[stable/polaris]: bump container version to 9.2.0 (major)

maximee-leroyy:
**Why This PR?**
Polaris container image tag has not been updated since one year

**Changes**
* Bump the container image tag to the latest version
* Use pull policy IfNotPresent since we are using fixed versions

**Checklist:**


sudermanjr:
Thanks for your PR, however we do not recommend the "IfNotPresent" pullPolicy for security reasons. Additionally, we pin only the minor version of the app intentionally, so that folks can pull in the patch changes automatically.

maximee-leroyy:
Hello @sudermanjr, not pinning the full image tag is a security issue because if a maintainer at polaris has his account compromised, he could push a new tag version with malicious code and all clusters will automatically pull it because the pull policy is set to Always, what do you think?

Btw did you close my PR by mistake or did i do something wrong?

PR #1546, opened 2024-11-04
[stable/polaris] Added mergeConfig flag to helm chart

KristianTrifork:
**Why This PR?**
The flag "mergeConfig" is possible to use for the docker command, but is not possible in helm chart. This has caused us to have two ways of configuration our Polaris, for local checks, we can use a custom config file, and use the 'mergeConfig' flag, while our deployment, needs the full configuration override.

Fixes #

**Changes**

Add the flag to the helm chart, by default setting it to false, so that it's only set if deliberately set to true.
This also ensures that the helm-chart deployment, and manual deployment are consistent

**Checklist:**

### DIFF
--- a/stable/polaris/Chart.yaml
+++ b/stable/polaris/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Validation of best practices in your Kubernetes clusters
 name: polaris
-version: 5.17.1
+version: 5.18.0
 appVersion: "8.5"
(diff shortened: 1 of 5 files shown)


vitorvezani (approved):
LGTM! Thanks for the contribution

sudermanjr (approved):
Thanks!

KristianTrifork:
Appreciate a merge whenever convenient, as we're looking to use this feature in our deployment.

Thanks.

PR #1588, opened 2025-05-06
[stable/goldilocks] fix: wrong `topologySpreadConstraints` helm value being used for dashboard

mirrajabi:
**Why This PR?**

Simple fix for the wrong `controller.topologySpreadConstraints` value being used instead of `dashboard.topologySpreadConstraints`.

**Checklist:**

### DIFF
--- a/stable/goldilocks/Chart.yaml
+++ b/stable/goldilocks/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 appVersion: "v4.13.0"
-version: 9.0.1
+version: 9.0.2
 kubeVersion: ">= 1.22.0-0"
 description: |
   A Helm chart for running Fairwinds Goldilocks. See https://github.com/FairwindsOps/goldilocks
--- a/stable/goldilocks/ci/pre-test-script.sh
+++ b/stable/goldilocks/ci/pre-test-script.sh
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+# This will label the first ndoe that is returned by kubectl get nodes so that we can test
+# the topologySpreadConstraints with `topologyKey: topology.kubernetes.io/zone`
+kubectl label $(kubectl get nodes -o name | head -n 1) topology.kubernetes.io/zone=test-zone-a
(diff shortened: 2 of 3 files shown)


sudermanjr (approved):
Good catch, thanks!

sudermanjr:
Change seems to be breaking the tests. 

```
Tolerations:         node.kubernetes.io/not-ready:NoExecute op=Exists for 300s
               node.kubernetes.io/unreachable:NoExecute op=Exists for 300s
Topology Spread Constraints: topology.kubernetes.io/zone:DoNotSchedule when max skew 1 is exceeded for selector app=alertmanager
Events:
 Type   Reason      Age  From        Message
 ----   ------      ---- ----        -------
 Warning FailedScheduling 23s  default-scheduler 0/1 nodes are available: 1 node(s) didn't match pod topology spread constraints (missing required label). preemption: 0/1 nodes are available: 1 Preemption is not helpful for scheduling.
 ```
 
 Might need to change the ci/test-values

mirrajabi:
Yeah so previously `controller.topologySpreadConstraints` in `test-values.yaml` was absent so it was working fine. But now that it is using the `dashboard.topologySpreadConstraints` which is present, it is failing.

Assuming the `topology.kubernetes.io/zone` label isn't configured properly on the nodes in the test cluster, we could either:
- add those (mocked) labels to nodes; Seeing the tests are ran using Kind (if I'm not mistaken), we can do that using [this configuration](https://kind.sigs.k8s.io/docs/user/configuration/#extra-labels).
- label the node after it's created inside the `fleet-install-test.sh` (I think this is the easiest option if it works)
- use another label as `topologyKey`
- use `whenUnsatisfiable: ScheduleAnyway` 🤷 
- remove the whole test 🤷 

Do you have a preference?

sudermanjr:
I would prefer not to modify the test cluster, just because that configuration is used across all the charts, and I have no clue what else we might break.

fleet-install-test.sh is not used by this chart, so that doesn't work.

Looks you can add a `ci/pre-test-script.sh` though, and that would allow labeling the nodes.

mirrajabi:
Thanks for the pointer. Using the pre script did it.

PR #1648, opened 2025-08-15
Change kubectl image from bitnami to alpine repository

vitorvezani:
**Why This PR?**
[internal INS-1296] change kubectl image from `bitnami` to `rancher` repository

Fixes #

**Changes**

*
*

**Checklist:**

### DIFF
--- a/stable/fairwinds-insights/values.yaml
+++ b/stable/fairwinds-insights/values.yaml
@@ -638,7 +638,7 @@ migrateHealthScoreJob:
 cronjobExecutor:
   image:
     repository: rancher/kubectl
-    tag: "v1.31.12"
+    tag: "v1.33.4"
   resources:
     limits:
       cpu: 100m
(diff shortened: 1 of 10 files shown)


stable/fairwinds-insights/values.yaml
```diff
@@ -637,8 +637,8 @@ migrateHealthScoreJob:
 
 cronjobExecutor:
   image:
-    repository: bitnami/kubectl
-    tag: "1.22.8"
+    repository: rancher/kubectl
```
jdesouza: What about alpine/kubectl?

vitorvezani: I chose `rancher` because they are maintaining their images since 1.17 (in case we need legacy versions)
`alpine/kubectl` only have 1.33.3 tag, looks like they just started maintaining kubectl images, and maybe they are not stable atm

jdesouza (approved):

stable/fairwinds-insights/values.yaml
vitorvezani: End-up using alpine due to rancher image being based on scratch

sudermanjr (approved):

PR #1697, opened 2025-11-05
[stable/vpa] support cert-manager ca bundle for webhook test

### LINKED ISSUE #1696: [stable/vpa] Property to disable webhook CA test
To start: thanks for the VPA chart!  It's unfortunate that one is not provided upstream.

With the VPA chart we do not want to use the patched keygen image that is now a few years old.  Instead we use cert-manager to generate the cert and inject into the admission webhook.  This works fine.
Unfortunately, the Helm test in the vpa chart is automatically enabled if the admission webhook is enabled, and it uses the Secret CA property [created by cert-keygen](https://github.com/kubernetes/ingress-nginx/blob/main/images/kube-webhook-certgen/rootfs/pkg/k8s/k8s.go#L250): `ca`.  But cert-manager uses the property `ca.crt` to store the CA cert.

Option 1: Property in Helm values to disable the admission webhook Helm test.  Less desired, it's a nice test.
Option 2: Enhance the Helm test to accommodate property `ca` or `ca.crt`: a transparent "fix".
Option 3: Property in Helm values to change the Secret property, defaulting to `ca`.

Using the nginx patched cert-keygen image is an alternative but we'd prefer cert-manager.

**Additional context**
We can submit a patch if you provide your preference from the solution options.

hobti01:
**Why This PR?**
Allow webhook Helm test to succeed when cert-manager generated Secret is used as source for webhook CA

Fixes #1696

**Changes**

* Add fallback from Secret property `ca` to property `ca.crt`

**Checklist:**

### DIFF
--- a/stable/vpa/Chart.yaml
+++ b/stable/vpa/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: vpa
 description: A Helm chart for Kubernetes Vertical Pod Autoscaler
 type: application
-version: 4.9.0
+version: 4.10.0
 appVersion: 1.4.1
 maintainers:
   - name: sudermanjr
--- a/stable/vpa/templates/tests/webhook.yaml
+++ b/stable/vpa/templates/tests/webhook.yaml
@@ -50,6 +50,16 @@ spec:
             {{ include "vpa.webhook.secret" . }} \
             -o=jsonpath="{.data.ca}")
 
+          # Get CA bundle if using cert-manager
+          if [ -z "$SECRET_CABUNDLE" ]
+          then
+            SECRET_CABUNDLE=$(kubectl \
+              get secret \
+              -n {{ .Release.Namespace }} \
+              {{ include "vpa.webhook.secret" . }} \
+              -o=jsonpath="{.data.ca\.crt}")
+          fi
+
           # Get configured CA bundle
           WEBHOOK_CABUNDLE=$(kubectl \
             get mutatingwebhookconfigurations.admissionregistration.k8s.io \


sudermanjr:
I wonder if there's a way to add another ci/ test values file to cover this. I _think_ we might have cert-manager installed in the test environment....

sudermanjr:
If you want, we could use a pre-script to install cert-manager in the test environment and test this. Example: https://github.com/FairwindsOps/charts/blob/7db239d3cd637033f909f06513fe14bb53feb60e/stable/insights-agent/ci/pre-test-script.sh#L4

it's not blocking IMO, but could be nice

hobti01:
> If you want, we could use a pre-script to install cert-manager in the test environment and test this. Example:
> it's not blocking IMO, but could be nice

Yes, would be nice.  Unfortunately I don't understand your test setup well enough to add this.  I think you'd need:
* [Self-Signed Issuer](https://cert-manager.io/docs/configuration/selfsigned/) so that a cert can sign itself
* Certificate resource that uses the Self-Signed issuer to create the TLS Secret
```yaml
apiVersion: cert-manager.io/v1
kind: Issuer
metadata:
  name: vpa-selfsigned-issuer
spec:
  selfSigned: {}
---
{{- if .Values.admissionController.enabled -}}
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: {{ index (splitList "/" (index .Values.admissionController.mutatingWebhookConfiguration.annotations "cert-manager.io/inject-ca-from")) 1 }}
spec:
  dnsNames:
    {{/* dnsname matches the Admission Controller Service name in the upstream chart */}}
    {{- $dnsname := (printf "%s-webhook" (include "vpa.fullname" .)) }}
    - '{{ $dnsname }}'
    - '{{ $dnsname }}.{{ .Release.Namespace }}'
    - '{{ $dnsname }}.{{ .Release.Namespace }}.svc'
  issuerRef:
    kind: Issuer
    name: vpa-selfsigned-issuer
  duration: "8760h0m0s"
  secretName: {{ include "vpa.webhook.secret" . }}
  subject:
    organizations:
      - vpa-admission
{{- end -}}
```
* Additional set of vpa test-values.yaml
```yaml
admissionController:
  enabled: true
  generateCertificate: false
  mutatingWebhookConfiguration:
    annotations:
      # This annotation injects the CA bundle from the Certificate into the webhook
      cert-manager.io/inject-ca-from: "<release-namespace>/vpa-webhook"
  # There is no need to have the webhook register itself.
  # The MutatingWebhookConfiguration is deployed with the chart
  # when admissionController.enabled: true
  registerWebhook: false
  # The keys in the Certificate Secret to map in to the admission controller
  tlsSecretKeys:
    - key: ca.crt
      path: caCert.pem
    - key: tls.crt
      path: serverCert.pem
    - key: tls.key
      path: serverKey.pem
```

hobti01:
I think it's probably a different enhancement to add a new flag to use cert-manager issued certs instead of certgen :)

sudermanjr (approved):

PR #1577, opened 2025-03-20
[stable/vpa] Add Release Namespace to VPA

emerchalegre:
**Why This PR?**
I was trying to deploy the VPA with terraform using helm_release, but because there was no namespace in the template, the deployment was always being performed in the default namespace, it was not respecting the namespace that I was informing in the code.

Fixes #

**Changes**
Changes proposed in this pull request:

* Added namespace property to vpa template

**Checklist:**

* [x] I have included the name of the chart in the title of this PR in square brackets i.e. `[stable/goldilocks]`.
* [x] I have updated the chart version in `Chart.yaml` following Semantic Versioning.
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] Any new values have been added to the README for the Chart, or `helm-docs --sort-values-order=file` has been run for the charts that support it.

PR #1150, opened 2023-04-18
[stable/goldilocks] [stable/vpa] Add .Release.Namespace template option

davidblum:
**Why This PR?**
We use [Carvel's Kapp-Controller](https://carvel.dev/kapp-controller/) to deploy our helm charts. If there is template option for `namespace: {{ .Release.Namespace }}` in the charts and sub-charts, kapp-controller puts the helm deployment into the default namespace.

As I understand helm, this should be a noop for normal `helm install` usage.

Fixes #

**Changes**
Changes proposed in this pull request:

* Add `.Release.Namespace` to `goldilocks` and `vpa` charts

**Checklist:**

* [x] I have included the name of the chart in the title of this PR in square brackets i.e. `[stable/goldilocks]`.
* [x] I have updated the chart version in `Chart.yaml` following Semantic Versioning.
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] Any new values have been added to the README for the Chart, or `helm-docs --sort-values-order=file` has been run for the charts that support it.

PR #1713, opened 2025-12-03
[stable/vpa] Add the ability to define a custom namespace to deploy the objects into

Aeternitaas:
### Why This PR?

The VPA chart does not currently support the ability to deploy its associated Kubernetes objects (deployments, PDBs, etc.) into a custom namespace. It will, by default, deploy all resources to `default`. We are submitting this PR so that we can allow for users to designate a custom namespace for the following templated resources under the VPA chart:

* Admission Controller:
  * `admission-controller-deployment.yaml`
  * `admission-controller-pdb.yaml`
  * `admission-controller-service-account.yaml`
  * `admission-controller-service.yaml`
* Recommender:
  * `recommender-deployment.yaml`
  * `recommender-pdb.yaml`
  * `recommender-podmonitor.yaml`
  * `recommender-service-account.yaml`
* Updater:
  * `updater-deployment.yaml`
  * `updater-pdb.yaml`
  * `updater-podmonitor.yaml`
  * `updater-service-account.yaml`

### Notes:
* `{{ .Release.Namespace }}` should be backwards compatible as the value defaults to `default` if unset. No new values are being set.
* I've read through the code pretty extensively but I'm not a maintainer (just someone who is using this Helm chart). There are a few issues I've encountered so far, namely:
  * ClusterRoleBindings, oddly, reference `{{ Release.Namespace }}`. When the Namespace value is set, it then references _nothing_. 
https://github.com/FairwindsOps/charts/blob/4168b92dec85ad00a784dc151fe51f171374e5d2/stable/vpa/templates/clusterrolebindings.yaml#L101-L134

If any of these "issues" I've found are actually intended, please let me know! 

### Changes:

* Adds the ability to set the VPA's namespace to something other than `default` by referencing `.Release.Namespace` in Gomplate. This change affects the Admission Controller, Recommender, and Updater.

### Checklist:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,7 +3,7 @@
 ## Guidelines on Contributing
 
 * Lint your changes:
-  * `docker run --rm -it -v ${PWD}:/charts -w /charts quay.io/helmpack/chart-testing:v3.14.0 ct lint --chart-yaml-schema scripts/schema.yaml --chart-dirs incubator --chart-dirs stable`
+  * `docker run --rm -it -v ${PWD}:/charts -w /charts quay.io/helmpack/chart-testing:v3.14.0 ct lint --chart-yaml-schema scripts/schema.yaml --chart-dirs incubator --chart-dirs stable --config scripts/ct.yaml`
 * End-to-end test your changes:
   * Install [kind 0.7.0+](https://github.com/kubernetes-sigs/kind/releases)
   * Install [chart-testing (ct)](https://github.com/helm/chart-testing/releases)
--- a/stable/vpa/Chart.yaml
+++ b/stable/vpa/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: vpa
 description: A Helm chart for Kubernetes Vertical Pod Autoscaler
 type: application
-version: 4.10.0
+version: 4.10.1
 appVersion: 1.4.1
 maintainers:
   - name: sudermanjr
--- a/stable/vpa/templates/admission-controller-deployment.yaml
+++ b/stable/vpa/templates/admission-controller-deployment.yaml
@@ -7,6 +7,7 @@ metadata:
   {{- .Values.admissionController.annotations | toYaml | nindent 4 }}
   {{- end }}
   name: {{ include "vpa.fullname" . }}-admission-controller
+  namespace: {{ .Release.Namespace }}
   labels:
     app.kubernetes.io/component: admission-controller
     {{- include "vpa.labels" . | nindent 4 }}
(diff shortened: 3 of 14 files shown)


CONTRIBUTING.md
```diff
@@ -3,7 +3,7 @@
 ## Guidelines on Contributing
 
 * Lint your changes:
-  * `docker run --rm -it -v ${PWD}:/charts -w /charts quay.io/helmpack/chart-testing:v3.14.0 ct lint --chart-yaml-schema scripts/schema.yaml --chart-dirs incubator --chart-dirs stable`
+  * `docker run --rm -it -v ${PWD}:/charts -w /charts quay.io/helmpack/chart-testing:v3.14.0 ct lint --chart-yaml-schema scripts/schema.yaml --chart-dirs incubator --chart-dirs stable --config scripts/ct.yaml`
```
Aeternitaas, line 6: This `docker ...` line just doesn't work unless you specify the `ct` config file, so I'm including this here :)

stable/vpa/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: vpa
 description: A Helm chart for Kubernetes Vertical Pod Autoscaler
 type: application
-version: 4.10.0
+version: 4.10.1
```
Aeternitaas, line 5: If this goes through, please let me know if this should be a major or minor release.

sudermanjr:
> The VPA chart does not currently support the ability to deploy its associated Kubernetes objects (deployments, PDBs, etc.) into a custom namespace.

This is not entirely accurate, as helm allows you to specify a namespace when installing, as do kubectl, and other methods of installation. Running a `helm template --namespace test-namespace` on this chart shows that the namespace field is blank, which means all objects will defer to the tool applying them to the cluster.  I generally prefer to avoid hard-coding the namespace in chart templates for this reason.

Is there a specific reason that hard-coding the namespace is necessary in your use-case?

Aeternitaas:
Yep. At our organization, we rely on Kustomize's [helmChartInflator](https://kubectl.docs.kubernetes.io/references/kustomize/builtins/#_helmchartinflationgenerator_) to render our manifests for us. This operation generates a huge manifest for us and we rely namespaces being defined within the manifests. Because of this, the lack of a `namespace:` field for items such as the `vpa-recommender` deployment, our systems defaults to applying the manifest to the `default` namespace due to the field being blank. As such, we rely on `namespace:` being defined because we cannot define the scope of our `kubectl apply` command without affecting several hundred other resources.

I'm sure you're probably aware, but there have been multiple other PRs that kinda suggest that others are running into the same issue and there's an old PR for goldilocks (that has been reverted) that attempted to do the same.
- https://github.com/FairwindsOps/charts/pull/1577
- https://github.com/FairwindsOps/charts/pull/1369
- https://github.com/FairwindsOps/charts/pull/1150

sudermanjr (approved):

sudermanjr:
Can you please sign the CLA?

Aeternitaas:
@sudermanjr Signed, thanks for being patient, had to clear it with my legal team :)

PR #1720, opened 2025-12-15
[stable/goldilocks]: Add Gateway API HTTPRoute support for dashboard

younsl:
**Why This PR?**

Add [Gateway API](https://gateway-api.sigs.k8s.io/) HTTPRoute support for the Goldilocks dashboard. This allows users to expose the dashboard using Gateway API instead of Ingress, which works with Istio, Envoy Gateway, and other Gateway API controllers.

Fixes #

**Changes**

- Add dashboard.httpRoute section in values.yaml with support for parentRefs, hostnames, and matches
- Add dashboard-httproute.yaml template for HTTPRoute resource
- Add mutual exclusion check to prevent enabling both Ingress and HTTPRoute simultaneously
- Update NOTES.txt to display HTTPRoute hostnames
- Update README.md with httpRoute values documentation

**Checklist:**

### DIFF
--- a/stable/goldilocks/templates/dashboard-httproute.yaml
+++ b/stable/goldilocks/templates/dashboard-httproute.yaml
@@ -1,6 +1,6 @@
 {{- if and .Values.dashboard.enabled .Values.dashboard.httpRoute.enabled }}
 {{- if .Values.dashboard.ingress.enabled }}
-{{- fail "Ingress and HTTPRoute resources cannot be enabled at the same time" }}
+{{- warn "Both Ingress and HTTPRoute are enabled. Consider disabling one after migration." }}
 {{- end }}
 {{- $fullName := include "goldilocks.fullname" . -}}
 apiVersion: gateway.networking.k8s.io/v1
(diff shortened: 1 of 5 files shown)


stable/goldilocks/templates/dashboard-httproute.yaml
```diff
@@ -0,0 +1,39 @@
+{{- if and .Values.dashboard.enabled .Values.dashboard.httpRoute.enabled }}
+{{- if .Values.dashboard.ingress.enabled }}
+{{- fail "Ingress and HTTPRoute resources cannot be enabled at the same time" }}
+{{- end }}
```
sudermanjr: I don't think this is really necessary. In fact some folks might want to have both for a time while they're migrating. Is there a `warn` option instead?

younsl: @sudermanjr Makes sense. Updated to use `warn` instead of `fail` so users can run both Ingress and HTTPRoute during migration. https://github.com/FairwindsOps/charts/pull/1720/commits/13fe3caf8c98898ee29f185a556579fb6f46262b

younsl:
@sudermanjr Removed the [`fail`](https://helm.sh/docs/chart_template_guide/function_list/#fail) validation to allow both Ingress and HTTPRoute to coexist during migration, similar to opencost/opencost-helm-chart#322.

younsl:
@sudermanjr PTAL

sudermanjr (approved):
LGTM. Thanks!

PR #1775, opened 2026-03-10
[stable/fairwinds-insights] INS-2037: Reduce temporal vulnerabilities

jdesouza:
**Why This PR?**

Fixes #

**Changes**

*
*

**Checklist:**


stable/fairwinds-insights/values.yaml
```diff
@@ -1010,6 +1010,8 @@ temporal:
   postgresql:
     enabled: true
   server:
+    image:
+      tag: "1.29.4"
```
vitorvezani, line 1014: I'm not a fan of overriding sub-charts version as we can:
- forget to update as chart evolves
- end-up with incompatible chart/images

stable/fairwinds-insights/values.yaml
```diff
@@ -969,7 +969,7 @@ temporalSchedulers: []
 temporalSchedulersJob:
   image:
     repository: temporalio/admin-tools
-    tag: "1.29.1-tctl-1.18.4-cli-1.5"
+    tag: "1.29.4"
```
vitorvezani, line 972: This one is out-side of their chart, so I guess it makes sense to update in case there is newer available images...

PR #1799, opened 2026-04-15
[stable/fairwinds-insights] INS-2205: Move timescale to CNPG

jdesouza:
BREAKING CHANGE: this will delete and replace any ephemeral timescale installations that are managed by this chart

Why This PR?
The old method of installing timescale was brittle and didn't work all
that well. The CNPG operator is amazing.

Fixes #

**Changes**

*
*

**Checklist:**


stable/fairwinds-insights/templates/postgresql-cluster.yaml
```diff
@@ -226,7 +212,7 @@ spec:
                   secret:
                     name: {{ .Values.postgresql.auth.existingSecret }}
               superuserSecret:
-                name: {{ .Values.postgresql.auth.existingSecret }}
+                name: {{ .Values.postgresql.auth.existingSuperUserSecret }}
```
vitorvezani: This seems to be a unrelated breaking change, should it be documented somewhere?

stable/fairwinds-insights/templates/postgresql-cluster.yaml
```diff
@@ -84,7 +85,6 @@ subjects:
   name: {{ .Release.Name }}-pg-creator
   namespace: {{ .Release.Namespace }}
 ---
-# Case 1: Operator is being installed, so wait for it and create cluster
 apiVersion: batch/v1
 kind: Job
```
vitorvezani, line 89: Do you think a single job for both is better than separating them?

stable/fairwinds-insights/templates/postgresql-cluster.yaml
```diff
@@ -294,3 +362,64 @@ spec:
...
+  labels:
+    {{- include "fairwinds-insights.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook-delete-policy": "before-hook-creation"
+spec:
+  images:
+    - major: {{ .Values.timescale.imageCatalog.major }}
+      image: {{ include "fairwinds-insights.timescaleImage" . }}
+---
+apiVersion: postgresql.cnpg.io/v1
```
vitorvezani: this code and the job above seems to be the same, maybe we can extract it and include there and here instead and have one source of truth to avoid drift

stable/fairwinds-insights/templates/postgresql-cluster.yaml
jdesouza: Not sure

jdesouza: Maybe we should simple and see if we see advantage or  need to split

stable/fairwinds-insights/templates/postgresql-cluster.yaml
jdesouza: done

stable/fairwinds-insights/templates/postgresql-cluster.yaml
jdesouza: done

vitorvezani (approved):

PR #1802, opened 2026-04-17
INS-2221 - add insights-mcp components (disabled by default)

vitorvezani:
**Why This PR?**
- deployment
- HPA
- PDB
- service
- ingress

Fixes #

**Changes**

*
*

**Checklist:**


stable/fairwinds-insights/templates/ingress.yaml
```diff
@@ -47,6 +47,12 @@ spec:
               serviceName: {{ $fullName }}-open-api
               servicePort: http
           {{- end }}
+          {{- if $.Values.mcp.ingress.enabled }}
```
jdesouza: Maybe this should check for mcp.enabled as well

stable/fairwinds-insights/values.yaml
```diff
@@ -423,6 +423,147 @@ openApi:
...
 
+# -- [Fairwinds Insights MCP server](https://quay.io/repository/fairwinds/insights-mcp-server) — Model Context Protocol bridge to the Insights API. **Disabled by default.**Transport `MCP_SERVER_TYPE` must be `sse` or `streamable` in-cluster (`stdio` is for local clients). Env vars and behavior follow the upstream server; the image defaults include `FAIRWINDS_API_BASE_URL` (override with `mcp.fairwindsApiBaseUrl` or rely on `options.host` / `ingress.hostedZones`). When `ingress.enabled` is true, `/mcp` is routed on the main Ingress to the MCP service.
+mcp:
+  # -- Deploy the Insights MCP server workload
+  enabled: false
+  image:
+    # -- Container image repository
+    repository: quay.io/fairwinds/insights-mcp-server
+    # -- Image tag (`latest` matches the published default; pin a digest or version for production)
+    tag: latest
```
jdesouza, line 434: do we want to use latest here? even though I know it is ours

stable/fairwinds-insights/values.yaml
```diff
@@ -423,6 +423,147 @@ openApi:
...
+  # -- Mount extra volumes (e.g. custom CA)
+  extraVolumeMounts: []
+  extraVolumes: []
+  # -- Optional init containers
+  initContainers: []
+  pdb:
+    # -- PodDisruptionBudget for the MCP Deployment
+    enabled: false
+    # -- Minimum available pods (map to PDB `minAvailable`)
+    minReplicas: 1
```
jdesouza, line 497: should not this be minAvailable ?

vitorvezani: It could but all other components use `*.pdb.minReplicas` so I kept the pattern

stable/fairwinds-insights/values.yaml
vitorvezani: that is the only one available, there is no proper release process just yet, we can revisit it once that is working

stable/fairwinds-insights/templates/ingress.yaml
vitorvezani: yea, I did that initially and removed because other components doesn't do it, but I think it makes sense here to be extra careful

vitorvezani: done

jdesouza (approved):

